PR #3, opened 2015-01-05
password reset and encrypting passwords with SHA256

fjoubert:
Please review and merge @theunsa

### DIFF
--- a/app/admin/admin.js
+++ b/app/admin/admin.js
@@ -157,7 +157,6 @@
                 })
                 .then(function (answer) {
                     passwordHash = CryptoJS.SHA256(answer).toString();
-                    console.log('new password\'s MD5 hash: ' + passwordHash);
 
                     UserService.resetPassword(user, passwordHash).then(function (result) {
                         console.log('reset password requested, result: ');
--- a/app/admin/admin.js
+++ b/app/admin/admin.js
@@ -2,7 +2,7 @@
     angular.module('katGui.admin', ['katGui.services', 'katGui.util'])
         .controller('AdminCtrl', AdminCtrl);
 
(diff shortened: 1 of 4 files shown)


app/admin/admin.js
```diff
@@ -69,7 +69,7 @@
                 id: user.id,
                 name: user.name,
                 email: user.email,
-                password: '1234',
+                password: CryptoJS.SHA256('password').toString(),
```
theunsa: string? shouldn't this just be the password variable tied to resetPasswordInput?

app/admin/admin.js
```diff
@@ -124,13 +124,55 @@
...
+                    "<div layout='row' layout-align='end' style='margin-top: 8px; margin-right: 8px; margin-bottom: 8px;'>" +
+                    "<md-button style='margin-left: 8px;' md-theme='{{themePrimaryButtons}}' aria-label='Cancel Reset' ng-click='cancel()'>Cancel</md-button>" +
+                    "<md-button style='margin-left: 8px;' md-theme='{{themePrimaryButtons}}' class='md-primary' aria-label='Reset Password' ng-click='answer(password)'><span>Reset</span></md-button>" +
+                    "</div>" +
+                    "</md-content></md-dialog>",
+                    targetEvent: event
+                })
+                .then(function (answer) {
+                    passwordHash = CryptoJS.SHA256(answer).toString();
+                    console.log('new password\'s MD5 hash: ' + passwordHash);
```
theunsa: console log MD5 instead of SHA256

app/admin/admin.js
fjoubert: This is the default password for now when creating a user. Resetting it gives you the opportunity to change it. We should probably create the default password on the the server when creating a user?

PR #9, opened 2015-02-20
Receptor status

fjoubert:


app/d3/d3-treemap.js
```diff
@@ -651,4 +886,4 @@ angular.module('katGui.d3')
                 });
             }
         };
-    });
\ No newline at end of file
+    });
```
theunsa: General comment: 
A lot of this code needs some refactoring and also documentation. 
Because we are grinding for the end Feb2015 deliverable this is understandable but we should not that we are accumulating technical debt here. For now, the fact that it works is enough but we should keep future maintainability in mind.

PR #25, opened 2015-03-05
added tasklog button in schedule details view

fjoubert:
@theunsa please merge

also:
- fixed md-select icon appearing in the wrong place when using a longer md-select than default
- changed service logging messages to indicate when they are connected and authorising

### DIFF
--- a/app/app.less
+++ b/app/app.less
@@ -551,3 +551,9 @@ md-toast {
     opacity: 1;
     max-height: 40px;
 }
+
+.md-select-label {
+    span:not(.md-select-icon) {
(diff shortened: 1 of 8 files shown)


app/app.less
```diff
@@ -551,3 +551,9 @@ md-toast {
     opacity: 1;
     max-height: 40px;
 }
+
+.md-select-label {
+    span:not(.md-select-icon) {
+        flex: 1;
+    }
+}
```
theunsa: Newline at end of file (ha ... github actually picks it up)

PR #26, opened 2015-03-06
added View Tasklog buttons to schedule block drafts and draft assignment displays

fjoubert:
@theunsa


app/scheduler/subarrays-draft-assignment/subarrays-draft-assignment.js
```diff
@@ -63,6 +63,12 @@
             }
         };
 
+        vm.viewSBTasklog = function () {
+            var sb = ObservationScheduleService.scheduleDraftData[vm.currentActionsMenuIndex];
+            ObservationScheduleService.viewTaskLogForSBIdCode(sb.id_code);
```
theunsa: How does this function know about 'sb' variable?

PR #30, opened 2015-03-18
More tests

fjoubert:
A whole (still incomplete and wip) unit test suite for existing controllers, functions, services, etc.
Fixes the issue we saw on the client's monitor-service that updates the receptor status display and operator control dsplay


theunsa:
Ok, this is a wild one:
- Only looked at the fixes you've made, its fine
- Did not go through all of these unit tests but can see that they are identifying problems which is good

theunsa:
@fjoubert: There are some merge conflicts ... please resolve those first.

fjoubert:
fixed the conflict in the package.json file

PR #40, opened 2015-05-14
Scheduler rework

fjoubert:
@theunsa, please merge.

Bringing katgui in line with katportal subarray_n changes


theunsa:
You should also note that right after marking a subarray as active ... that subarray will report state 'initialising' and it will remain there for a few minutes until all initialising tasks are done, after which it will report 'active'.

So during that initialisation time the GUI should display some kind of wait/progress indicator.

fjoubert:
I'll note that down as a todo, thanks

PR #45, opened 2015-05-21
Sensor list and other changes

fjoubert:
Included in this pull request:

* New sensor list display
* Changed ':' sensor name separators to '.' and all the fixes that goes with that on monitor connections
* Added prefixes to subscribes to be in line with server changes
* New version of underscorejs
* Added draggable and resizable directives
* Optimised sensor graph d3 class - needs more work for dot-hover data points - wip

### DIFF
--- a/app/sensor-graph/sensor-graph.js
+++ b/app/sensor-graph/sensor-graph.js
@@ -180,7 +180,7 @@
                         for (var i = 0; i < result[attr].length; i++) {
                             newData.push({
                                 Sensor: attr,
-                                ReceivedTimestamp: result[attr][i][0],
+                                Timestamp: result[attr][i][0],
                                 Value: result[attr][i][1],
                                 Details: sensor
                             });
@@ -234,7 +234,7 @@
                         for (var i = 0; i < result[attr].length; i++) {
                             newData.push({
                                 Sensor: attr,
-                                ReceivedTimestamp: result[attr][i].sample_ts,
+                                Timestamp: result[attr][i].sample_ts,
                                 Value: result[attr][i].value
                             });
                         }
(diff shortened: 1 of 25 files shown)


app/sensor-graph/sensor-graph.js
```diff
@@ -179,7 +180,7 @@
                         for (var i = 0; i < result[attr].length; i++) {
                             newData.push({
                                 Sensor: attr,
-                                Timestamp: result[attr][i][0],
+                                ReceivedTimestamp: result[attr][i][0],
```
theunsa: Need to think about this ... on sensor-graph the user will be interested in the actual device sample time of every sample and not really the received-timestamp which represents when katportal's resource connection got an update.

I see only the name changed and not the index reference so you might actually still be using the real sample timestamp.

fjoubert: Sure, the wording is a bit misleading. Will need to change it to something that makes more sense. It was a quick fix to make the multi line chart work after changes were made for it for sensor-list.

Because we only get 2 values (sample_ts and value) from katstore-webserver, we still plot the correct values. This is not good enough when connecting a live data feed to the graph though. Sensor graph is definitely still very much WIP, this is one of the many things I need to touch on.

Thanks for the comment though.

theunsa: Ok, so I suggest you rename it back to Timestamp or SampleTimestamp to avoid confusion with ReceiveTimestamp which is actually a different timestamp value in the KATCP context.

fjoubert: Fair points, renamed in 2282ec6

PR #48, opened 2015-05-25
Sensor list updates

fjoubert:
@theunsa please review and merge

added resize convenience buttons to graph in sensor list
fixed sensor list ordering, added config options for strategy types

note: next pull request massively refactors config display

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -103,7 +103,7 @@
             $rootScope.sensorListStrategyType = 'event-rate';
         }
         if (!$rootScope.sensorListStrategyInterval) {
-            $rootScope.sensorListStrategyType = 3;
+            $rootScope.sensorListStrategyInterval = 3;
         }
 
         vm.currentUser = null;
--- a/app/app.js
+++ b/app/app.js
@@ -96,6 +96,16 @@
             $rootScope.showNavbar = value;
         });
 
+        $rootScope.sensorListStrategyType = $localStorage['sensorListStrategyType'];
+        $rootScope.sensorListStrategyInterval = $localStorage['sensorListStrategyInterval'];
+
+        if (!$rootScope.sensorListStrategyType) {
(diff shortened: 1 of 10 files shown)


app/app.js
```diff
@@ -96,6 +96,16 @@
...
         });
 
+        $rootScope.sensorListStrategyType = $localStorage['sensorListStrategyType'];
+        $rootScope.sensorListStrategyInterval = $localStorage['sensorListStrategyInterval'];
+
+        if (!$rootScope.sensorListStrategyType) {
+            $rootScope.sensorListStrategyType = 'event-rate';
+        }
+        if (!$rootScope.sensorListStrategyInterval) {
+            $rootScope.sensorListStrategyType = 3;
```
theunsa: I guess Interval is meant here ...

fjoubert: Good catch, yes

app/services/sensors-service.js
```diff
@@ -159,7 +159,16 @@
...
-            api.sendSensorsCommand('set_sensor_strategy', [guid, resource_name, '', 'event-rate', 3, 3]);
+            api.sendSensorsCommand(
+                'set_sensor_strategy',
+                [
+                    guid,
+                    resource_name,
+                    '',
+                    $rootScope.sensorListStrategyType,
+                    $rootScope.sensorListStrategyInterval,
+                    $rootScope.sensorListStrategyInterval
```
theunsa: Why is Interval sent twice?

fjoubert: Event-rate strategy type requires a minimum and a maximum interval, which for now, we set as the same value. The second interval is ignored by the webserver for other strategy types where it is not applicable, like 'period' and in the case of 'event' both parameters are ignored.

theunsa: Got it.

PR #49, opened 2015-05-25
Display date

fjoubert:
@theunsa please merge

* Massively refactored config display
* Added day of year and date displays

note: this was branched after the 'Sensor list updates' pull request, so maybe look at that one first


fjoubert:
I see I also have the fixes for ordering sensor list stuff in this pull request, probably shouldve been in its own PR

PR #51, opened 2015-05-26
listen to correct event when routing to avoid infinite looping bug

fjoubert:


fjoubert:
Merged own pull request for a one liner fix.

PR #58, opened 2015-06-09
Central logger link

fjoubert:
@theunsa please merge

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -54,14 +54,13 @@
                 secondary: 'dark-secondary',
                 primaryButtons: 'dark-buttons'
             }])
-        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devf.camlab.kat.ac.za' : window.location.origin)
         .config(configureKatGui)
         .run(runKatGui)
         .controller('ApplicationCtrl', ApplicationCtrl);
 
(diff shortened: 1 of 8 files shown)


app/app.js
```diff
@@ -54,6 +54,7 @@
                 secondary: 'dark-secondary',
                 primaryButtons: 'dark-buttons'
             }])
+        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devf.camlab.kat.ac.za' : window.location.origin)
```
theunsa: Note: We need to think about a better way to handle development especially now that other team members are going to start doing GUI development.

app/app.js
```diff
@@ -345,6 +346,11 @@
                 });
         };
 
+        vm.openCentralLogger = function () {
+            //TODO get from config and eventually redo central logger
+            KatGuiUtil.openRelativePath('', 9021);
```
theunsa: Should this port number not also be a constant?

app/app.js
fjoubert: Yeah we really should. I am moving the constant back to where it was because it broke too many tests.

app/app.js
fjoubert: Good point, going to make it a constant.

fjoubert:
@theunsa please merge if you are satisfied, would like to get jenkins happy again for katgui :)

PR #61, opened 2015-06-19
Custom health display rework

fjoubert:
@theunsa please review and merge

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -115,6 +115,10 @@
             $rootScope.showAlarms = true;
         }
 
+        $rootScope.isNavbarVisible = function () {
+            return vm.showNavbar;
+        };
+
         $rootScope.showSimpleToast = function (message) {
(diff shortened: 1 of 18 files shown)


app/app.js
```diff
@@ -373,9 +378,12 @@
             $httpProvider.defaults.useXDomain = true;
             delete $httpProvider.defaults.headers.common['X-Requested-With'];
         }
-
+        //todo nginx needs the following config before we can switch on html5Mode
```
theunsa: mmh ... will have to add reminder for this somewhere

fjoubert: its on my never ending todo list :)

fjoubert: Oh, but I haven't activated html5 mode (which takes out the # in the url) because there is some issues that I need to figure out first. So no need to do it just yet.

PR #70, opened 2015-06-26
add a button to switch on and off flood lights

ofaletse:
@fjoubert May you please review.

### DIFF
--- a/app/operator-control/operator-control.html
+++ b/app/operator-control/operator-control.html
@@ -16,6 +16,7 @@
             <md-button ng-click="vm.resumeOperations()" class="md-raised md-primary">Resume Operations</md-button>
             <md-button disabled ng-click="shutdownComputing()" class="md-raised md-primary">Shutdown Computing
             </md-button>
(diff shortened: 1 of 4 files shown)


app/operator-control/operator-control.html
```diff
@@ -16,6 +16,7 @@
             <md-button ng-click="vm.resumeOperations()" class="md-raised md-primary">Resume Operations</md-button>
             <md-button disabled ng-click="shutdownComputing()" class="md-raised md-primary">Shutdown Computing
             </md-button>
+            <md-button ng-click="vm.toggleFloodLights()" class="md-raised md-primary">Switch Floodlights {{!vm.floodLightSensor.value? 'on' : 'off'}}</md-button>
```
theunsa: @lvdheever / @ofaletse : I have my doubts whether this should be on the OperatorControl display.

The floodlights are always related to the video system, so I would expect to find the toggling of the lights on the VDS Display. Else the user will now have to navigate two displays to do one thing.

lvdheever: Yeah - I think I agree with your suggestion. It was added to the Control display in previous KatGUI as it just was convenient there.

PR #87, opened 2015-07-06
Op ctrl rework

fjoubert:
@theunsa Control Service rework to use http endpoints instead of a websocket connection

Note: tests needs a go over again to get the build to pass


theunsa:
Will wait for https://github.com/ska-sa/katportal/pull/66 to be fixed and merged before tackling this.

PR #92, opened 2015-07-20
renamed all Sub-Array... to Subarray, added datetime formatting constant

fjoubert:
@theunsa please merge


fjoubert:
Self merge because of nature of changes

PR #93, opened 2015-07-20
changed login to POST instead of GET

lanceWilliams:
Changed login to POST instead of GET. Both login and logout are now POST but send no data, they use the session token instead.

### DIFF
--- a/app/services/session-service.js
+++ b/app/services/session-service.js
@@ -30,10 +30,7 @@
         api.login = function (session_id) {
 
             $rootScope.jwt = session_id;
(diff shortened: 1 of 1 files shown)


app/services/session-service.js
```diff
@@ -30,7 +30,10 @@
         api.login = function (session_id) {
 
             $rootScope.jwt = session_id;
-            $http(createRequest('get', urlBase + '/user/login'))
+            $http(createRequest('post', urlBase + '/user/login',
+                {
+                    role: ""
+                }))
```
fjoubert: What is the purpose of the empty role in the post body?

PR #97, opened 2015-07-22
Gui hotfix

lanceWilliams:
Urlbase for userlog-service might need to change for this to work. Not sure where to configure.


theunsa:
The problem with the current urlbase is that it points to a new path that will have to be configured on nginx. But as we've mentioned before ... this should be the same as the katcontrol webserver path because the userlogs should not have its own web server.

PR #122, opened 2015-08-07
use receptor name instead receptor number

ofaletse:
@fjoubert please review


lvdheever:
This will not work for KAT-7 :-(

PR #143, opened 2015-08-25
added NASA APOD picture and explanation widget for dashboard

fjoubert:
@lvdheever, I was looking for something fun to do, so I added an APOD widget to the dashboard :)


theunsa:
:+1: Cool!

PR #154, opened 2015-10-29
Graphic receptor status

fjoubert:
some improvements after feedback

screen shot 2015-10-29 at 12 45 10 pm


brickZA:
I like how tired the 'stop' mode receptors look, like they've gone to sleep
;)

On Thu, Oct 29, 2015 at 12:45 PM, Francois Joubert <notifications@github.com

fjoubert:
Added force layouts for receptor health
Added a force layout per subarray - experimental

PR #157, opened 2015-12-03
CB-493 use bands and products from system config

fjoubert:
merge with https://github.com/ska-sa/katconfig/pull/104

### DIFF
--- a/app/services/control-service.js
+++ b/app/services/control-service.js
@@ -1,7 +1,7 @@
 (function () {
 
     angular.module('katGui.services')
-        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devq.camlab.kat.ac.za' : window.location.origin)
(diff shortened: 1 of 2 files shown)


app/services/control-service.js
```diff
@@ -1,7 +1,7 @@
 (function () {
 
     angular.module('katGui.services')
-        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devq.camlab.kat.ac.za' : window.location.origin)
+        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devf.camlab.kat.ac.za' : window.location.origin)
         .service('ControlService', ControlService);
```
lvdheever: Not sure what this URL is for, but it looks a bit wrong to have a devX site hardcoded in here.

lvdheever:
Merging even if there is a comment on the hardcoded devf

app/services/control-service.js
fjoubert: This is specific to running the gui on localhost for development, it's where you change where your localhost is connecting to a dev system. Haven't been able to find an elegant solution for this yet.

PR #158, opened 2015-12-03
CB-493 removed all traces of ap_ sensors from katgui and use proper pos_ sen…

fjoubert:
…sors for pointing displays

### DIFF
--- a/app/d3/receptor-health-directives-spec.js
+++ b/app/d3/receptor-health-directives-spec.js
@@ -25,7 +25,7 @@ describe('Directive: d3-chart', function () {
                             sensor: 'sensors_ok',
                             children: [{
                                 name: 'ap',
-                                sensor: 'ap_device_status',
+                                sensor: 'device_status',
                                 children: [
                                     {
                                         name: 'bp',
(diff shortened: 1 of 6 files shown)


app/d3/receptor-health-directives-spec.js
```diff
@@ -71,7 +71,7 @@ describe('Directive: d3-chart', function () {
         scope.$digest();
         document.body.appendChild(element[0]);
         expect(angular.element(document.querySelector('#m011_sensors_ok')).attr('class')).toContain('inactive-child child');
-        expect(angular.element(document.querySelector('#m011_ap_device_status')).attr('class')).toContain('inactive-child');
+        expect(angular.element(document.querySelector('#m011_device_status')).attr('class')).toContain('inactive-child');
     });
```
lvdheever: Unfortunately the proxies don't implement device-status on this level yet. Best "generic" sensor that can be used for now is "m0xx_all_ok".

fjoubert: That doesn't matter at all, this is a unit test, so the actual sensor name is irrelevant :)

PR #162, opened 2015-12-07
Improved UX for custom health displays; fixes an issue with resizing on top health

fjoubert:


fjoubert:
also added a view task log button for completed schedule blocks in the detailed subarray display

PR #174, opened 2016-01-15
Karoocamv10

fjoubert:
same as #173 just for kat7


fjoubert:
Closing PR because it was opened incorrectly

PR #204, opened 2016-03-16
userlogs rework number 2 for kat7

fjoubert:
see https://skaafrica.atlassian.net/browse/CB-854 for details on changes


fjoubert:
self merge

PR #216, opened 2016-05-09
draw max/min lines when there are values for it; allow multiple discr…

fjoubert:
…ete sensors to be plotted on the same graph; upgrade to material 1.0.8


martinslabber:
I am happy think @bxaia  should have the final say.

PR #220, opened 2016-06-06
fixed config for instrumental config and display a good error message…

fjoubert:
… when sensor groups failed


fjoubert:
Fixes https://skaafrica.atlassian.net/browse/CB-1129

PR #221, opened 2016-06-06
add tail of nodeman process log on process-control display

fjoubert:


fjoubert:
requires https://github.com/ska-sa/katportal/pull/200

PR #224, opened 2016-06-13
Refactored a lot of scheduler for better UX

fjoubert:


fjoubert:
Also improved sensor list display layout and process control display layout.

fjoubert:
Related JIRA https://skaafrica.atlassian.net/browse/CB-1063

PR #225, opened 2016-06-13
maintenance and faulty buttons on receptor status display

fjoubert:
Related JIRA - https://skaafrica.atlassian.net/browse/CB-1060


fjoubert:
Merge after https://github.com/ska-sa/katgui/pull/224

PR #229, opened 2016-06-17
KATGui receptors in maintenance and faulty to log in userlogs

bxaia:
@fjoubert please review

### DIFF
--- a/app/scheduler/scheduler-home.js
+++ b/app/scheduler/scheduler-home.js
@@ -35,7 +35,7 @@
             }
         });
 
-	UserLogService.listTags()
+        UserLogService.listTags();
 
         ConfigService.getSystemConfig()
             .then(function (systemConfig) {
--- a/app/scheduler/scheduler-home.js
+++ b/app/scheduler/scheduler-home.js
@@ -11,7 +11,7 @@
         .controller('SchedulerHomeCtrl', SchedulerHomeCtrl);
 
     function SchedulerHomeCtrl($state, $rootScope, $scope, $interval, $log, SensorsService, ObsSchedService, $timeout, KatGuiUtil,
-                               NotifyService, MonitorService, ConfigService, $stateParams, $q, $mdDialog, UserService) {
+                               UserLogService, NotifyService, MonitorService, ConfigService, $stateParams, $q, $mdDialog, UserService) {
 
         var vm = this;
         vm.childStateShowing = $state.current.name !== 'scheduler';
@@ -34,6 +34,8 @@
                 vm.users.push(UserService.users[i]);
             }
         });
+
+	UserLogService.listTags()
 
         ConfigService.getSystemConfig()
             .then(function (systemConfig) {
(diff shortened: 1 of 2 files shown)


app/scheduler/scheduler-home.js
```diff
@@ -35,6 +35,8 @@
             }
         });
 
+	UserLogService.listTags()
```
fjoubert: Indentation is strange here and theres no ';' at the end of the line. Go set 'soft tabs' and tab length in atom preferences -> settings.

 Try installing a js linter in atom to show you these problems as you work. Open atom -> preferences -> packages -> search for jshint or something similar. Code style and good habits are extremely important in a dynamic language. Especially when we minimise code - missing ';' could cause lots of pain.

app/services/obs-sched-service.js
```diff
@@ -64,10 +65,40 @@
 
         api.markResourceFaulty = function (resource, faulty) {
             api.handleRequestResponse($http(createRequest('post', urlBase + '/resource/' + resource + '/faulty/' + faulty)));
-        };
+	    var tags = [];
+	    tags.push(_.find(UserLogService.tags, function(item){return item.name === 'status'}));
+	    var tagresource = _.find(UserLogService.tags, function(item){return item.name === resource});
+	    if (tagresource) {
+		tags.push();
```
fjoubert: you are not pushing anything to the array if the tagresource is defined, it should be `tags.push(tagresource);`. Also, change the variable to camelCase so that its tagResource.

Do the same kind of check of the 'status' tag, so that we do not stop the dialog from appearing when the tag does not exist.

app/services/obs-sched-service.js
```diff
@@ -64,10 +65,40 @@
...
+            UserLogService.editUserLog(newlog, true)
+	 };
 
         api.markResourceInMaintenance = function (resource, maintenance) {
             api.handleRequestResponse($http(createRequest('post', urlBase + '/resource/' + resource + '/maintenance/' + maintenance)));
+	    var tags = [];
+            tags.push(_.find(UserLogService.tags, function(item){return item.name === 'maintenance'}));
+            var tagresource = _.find(UserLogService.tags, function(item){return item.name === resource});
+            if (tagresource) {
+                tags.push();
```
fjoubert: same as above

app/services/obs-sched-service.js
```diff
@@ -64,10 +65,40 @@
...
+            }
+
+	    var newlog = {
```
fjoubert: please double check indentation and make this neat, indentation doesn't break code like in python, but it can cause headaches later on in development!

app/services/obs-sched-service.js
```diff
@@ -64,10 +65,40 @@
...
+                start_time: $rootScope.utcDateTime,
+                end_time: '',
+                tags: _.find(UserLogService.tags, function(item){return item.name === 'maintenance'}),
```
fjoubert: the tags needs to be an array!

PR #230, opened 2016-06-22
	refactored sensor graph interactive features to perform well enough up to at least 100000 data points

fjoubert:


fjoubert:
related jira - https://skaafrica.atlassian.net/browse/CB-1120

PR #232, opened 2016-06-30
Timeline display

fjoubert:
Added an activity display with custom sensor graphs, see screenshot for an example.

This is still a bit rough around the edges, it is the 'first iteration' as in https://skaafrica.atlassian.net/browse/CB-1115

screen shot 2016-06-30 at 3 32 15 pm


fjoubert:
The schedule block width gets bigger depending on the expected duration

lvdheever:
Not sure if it is in the making somewhere, but a scrolling window of the last 10 or so lines from the activity.log could be very useful somewhere. That will also make us give due attention to clearing it up and making it really usable (like removing the user.log stuff and using inputs from operators and ourselves to add/remove info as required)

martinslabber:
Hi Lize. Since Francois is busy organising a wedding I'll chirp in: He mentioned a few ideas around that similar to what you said, this is on his radar. We did cross the path that logging needs rework and there are some interesting industry tools for log display and analysis with good API's. So definitely bring this idea up when we chat about logging one day.

From what Francois showed me this is only an appetiser - a technology preview.

fjoubert:
As Martin says - this is the very first iteration of an activity display. I left out the actual activity log for now so that I could have a version we can use and that could be useful already.

Future ideas would include things like the activity log, a way to save and share a display template, a way to plot multiple sensors on the same graph and refinements to what we currently have.

This PR also includes some minor bug fixes for the sensor graph, which is what is used to display the sensor info on the activity display.

PR #233, opened 2016-07-15
Load user_product names, not the whole user_product:(cbf_product,sp_p…

profcalculus:
…roduct) map


fjoubert:
After a talk with Charles - this is working as intended at the moment and does not require this PR.

PR #235, opened 2016-07-29
CB-1301 - compress dist/ folder for upload

fjoubert:
After the compression we will have an artifact we can upload via jenkins.

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -14,3 +14,11 @@ npm install
 echo "## Performing gulp build ..."
 gulp build
 
+echo "## Compressing dist/ folder ..."
+GIT_REF=`git rev-parse --short HEAD`
+VERSION=`python -m json.tool dist/version.txt | grep -Po '\d\.\d\.\d'`
+DATE=`date +"%Y-%m-%d"`
+TIME=`date +"%H:%M"`
+FILENAME="katgui-$VERSION-$DATET$TTIME-$GIT_REF.tar.gz"
+tar -zcvf $FILENAME dist/
+echo "## KATGUI compressed to $FILENAME"
--- a/update.sh
+++ b/update.sh
@@ -15,10 +15,9 @@ echo "## Performing gulp build ..."
 gulp build
 
 echo "## Compressing dist/ folder ..."
-GIT_REF=`git rev-parse --short HEAD`
-VERSION=`python -m json.tool dist/version.txt | grep -Po '\d\.\d\.\d'`
-DATE=`date +"%Y-%m-%d"`
-TIME=`date +"%H.%M"`
-FILENAME="katgui-$VERSION-$DATE"T"$TIME-$GIT_REF.tar.gz"
+VERSION=`kat-get-version.py`
+FILENAME="katgui-$VERSION.tar.gz"
+TIMESTAMP=`date +%s`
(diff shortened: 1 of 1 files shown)


update.sh
```diff
@@ -14,3 +14,11 @@ npm install
 echo "## Performing gulp build ..."
 gulp build
 
+echo "## Compressing dist/ folder ..."
+GIT_REF=`git rev-parse --short HEAD`
+VERSION=`python -m json.tool dist/version.txt | grep -Po '\d\.\d\.\d'`
+DATE=`date +"%Y-%m-%d"`
+TIME=`date +"%H:%M"`
+FILENAME="katgui-$VERSION-$DATET$TTIME-$GIT_REF.tar.gz"
```
ajoubertza: Look like some typos here.  `$DATET` -> `$DATE`, and `$TTIME` -> `$TIME`

fjoubert: Woops - well spotted. I copied the wrong snippet 👎

update.sh
```diff
@@ -14,3 +14,11 @@ npm install
 echo "## Performing gulp build ..."
 gulp build
 
+echo "## Compressing dist/ folder ..."
+GIT_REF=`git rev-parse --short HEAD`
+VERSION=`python -m json.tool dist/version.txt | grep -Po '\d\.\d\.\d'`
```
ajoubertza: I'm not sure what is in version.txt, but will the version numbers always be single digits? E.g. no version 10.

update.sh
```diff
@@ -14,3 +14,11 @@ npm install
 echo "## Performing gulp build ..."
 gulp build
 
+echo "## Compressing dist/ folder ..."
+GIT_REF=`git rev-parse --short HEAD`
```
martinslabber: To get the version string please use 'kat-get-version.py' from the katversion package. This will allow us to be consistent over all our products. I know this is not a python package but we added the shell command to katversion for this reason.  I know it adds a dependency.

update.sh
```diff
@@ -14,3 +14,11 @@ npm install
...
+VERSION=`python -m json.tool dist/version.txt | grep -Po '\d\.\d\.\d'`
+DATE=`date +"%Y-%m-%d"`
+TIME=`date +"%H.%M"`
+FILENAME="katgui-$VERSION-$DATE"T"$TIME-$GIT_REF.tar.gz"
+tar -zcvf $FILENAME dist/
```
martinslabber: O how I have longed for this. 👍 

A note.  To keep with Jenkins and Debian style please do not version the file name. The build and the resulting package are versioned. What would be nice it to paste the version string into a file.

fjoubert:
These are all great ideas. Here's what the above gives us:

- a filename that looks something like this: `katgui-0.1.dev1225+fj.compress.dist.907c134.tar.gz`
- a version file in that compressed file that looks like this: `{"version":"0.1.dev1225+fj.compress.dist.907c134","buildDate":"1469782744000"}`

Note: we absolutely need this version.txt file for checking cached versions of the gui.

update.sh
fjoubert: Nice, I didn't know about the shell command :)

update.sh
fjoubert: A good point, one that is addressed in commit 907c134

PR #247, opened 2016-09-12
First version of the utilisation reports

fjoubert:
Merge with https://github.com/ska-sa/katportal/pull/231

This display shows the following (from devf data): 

screen shot 2016-09-12 at 1 41 22 pm
screen shot 2016-09-12 at 1 41 31 pm

Then the export pdf button will give you:
screen shot 2016-09-12 at 3 03 58 pm
screen shot 2016-09-12 at 3 04 05 pm
screen shot 2016-09-12 at 3 04 19 pm
screen shot 2016-09-12 at 3 04 23 pm

This is a quick and dirty and could probably benefit from some refactoring, but it is good enough for now because it works.

### DIFF
--- a/app/services/control-service.js
+++ b/app/services/control-service.js
@@ -1,7 +1,7 @@
 (function () {
 
     angular.module('katGui.services')
-        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://portal.mkat.devkaroo.camlab.kat.ac.za' : window.location.origin)
+        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devf.camlab.kat.ac.za' : window.location.origin)
         .service('ControlService', ControlService);
 
     function ControlService($rootScope, $http, SERVER_URL, NotifyService, KatGuiUtil) {
--- a/app/app.js
+++ b/app/app.js
@@ -312,7 +312,11 @@
             }
         };
         $rootScope.objectKeys = function (obj) {
-            return Object.keys(obj);
+            if (obj) {
+                return Object.keys(obj);
+            } else {
+                return [];
+            }
         };
         $rootScope.openCentralLogger = function () {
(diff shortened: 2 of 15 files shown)


lvdheever:
This looks like very good detail, but we will have to work on how it is presented to the user for easier consumption. I'm thinking a few tables with columns to collate this information to a better understandable format. 
Will find some time today to chat with you.

fjoubert:
Definitely could use a lot of refinement. But this is a good place to start off from seeing as all the data is captured, we can just work on formatting.

I could probably explain what you are seeing as 'not adding up'.

I am back on Thursday only, we can have a chat then.

fjoubert:
Screen shots for the report as it stands now:

screen shot 2016-09-19 at 11 21 01 am
screen shot 2016-09-19 at 11 21 17 am

fjoubert:
This PR is now ready to be reviewed and/or merged.

app/services/control-service.js
```diff
@@ -1,7 +1,7 @@
 (function () {
 
     angular.module('katGui.services')
-        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://portal.mkat.devkaroo.camlab.kat.ac.za' : window.location.origin)
+        .constant('SERVER_URL', window.location.host === 'localhost:8000' ? 'http://monctl.devf.camlab.kat.ac.za' : window.location.origin)
         .service('ControlService', ControlService);
```
lvdheever: Need to revert this?

fjoubert: I've been wanting to add something that allows me to remove this line so that we can host the gui locally in 'developer mode' or something like that. Something where you can enter a portal location when hosting locally. In fact - I just made a ticket for it: https://skaafrica.atlassian.net/browse/CB-1499. For now I'd like to keep the SERVER_URL constant pointing to devf.

PR #249, opened 2016-09-27
Development mode; sensor graph improvements; bug fixes

fjoubert:
This PR contains a bunch of changes:

* There is now a 'development mode' which allows you to specify the url of katportal that you would like to connect to, this replaces the constant we had to keep changing in the control-service.js. (NOTE: the development mode input is only visible when hosting the gui locally on port 8000)
* Some basic checks has been added to the login form see screenshot:

screen shot 2016-09-27 at 1 40 59 pm

* Bug fix - when trying to log in as lead operator when there is a lead operator, but then cancelling out of the dialog box, now correctly logs you in as read only. Also changed the wording on the dialog buttons when taking lead operator from someone to be clearer.
* Bug fix - reworked downloading csv data from sensor graph, you can now download LARGE sets of sensor samples (millions of samples), previously it would fail if there were too many samples.

* Numerous improvements to sensor graph rendering performance.
* Correctly remove the context path when removing a sensor path from sensor graph

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,20 +1,18 @@
 node('docker') {
     stage 'Cleanup workspace'
-    dir('katgui') {
-        deleteDir()
-    }
-    sh 'rm -rf *.deb'
+    sh 'chmod 777 -R .'
+    sh 'rm -rf *'
 
     docker.image('camguinode:latest').inside('-u root') {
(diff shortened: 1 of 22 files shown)


fjoubert:
Also fixed up the jenkinsfile to checkout the correct head of the branch.

PR #258, opened 2016-12-12
reworked scheduling to only allow approved sb's to be scheduled

fjoubert:
also updated message handling to be inline with new db trigger based notifications


fjoubert:
Please see https://skaafrica.atlassian.net/browse/CB-1679 for the related branches.

app/scheduler/schedule-block-drafts/schedule-block-drafts.html
```diff
@@ -72,13 +72,10 @@
                         <span class="icon-button fa fa-ellipsis-v" md-ink-ripple style="padding: 6px" ng-click="$mdOpenMenu(); $event.stopPropagation()"></span>
                         <md-menu-content>
                             <md-menu-item>
-                                <md-button ng-disabled="!$root.expertOrLO && $root.currentUser.req_role !== 'control_authority'" ng-click="item.editing = true">Edit Draft</md-button>
+                                <md-button ng-disabled="!parent.vm.iAmAtLeastCA()" ng-click="item.editing = true">Edit Draft</md-button>
```
martinslabber: Would this still be "Edit Draft" or should this also change to "Edit Approved".

fjoubert: No - I changed it to just `Edit` in 397029a.

PR #267, opened 2017-02-27
Added a scheduler mode dropdown to all scheduler displays with subarrays

fjoubert:
Branched from user/fjoubert/CB-1859/change-sb-pb-order-vars and to be merged after https://github.com/ska-sa/katgui/pull/266.


fjoubert:
Also in this PR: fix cam-components display to work with split proxies (@ajoubertza - just for you)

PR #268, opened 2017-02-28
fix incorrect userlogs endpoint; allow editing of userlogs for own userlogs from userlogs reports display

fjoubert:


lvdheever:
Great tx Francois!

PR #288, opened 2017-03-15
Correctly disable clicking when logging in as a CA and not allowed to control a subarray

fjoubert:
This change is necessary because a while back I changed some of the inline <md-button> elements to be <span>'s. ng-disabled only works with input type elements, so we have do some more work when disabling a span (i.e. prevent the click event to propagate).

Also: 
* Allow a CA to mark a resource as faulty when he is the CA of that subarray
* do not try to JSON.parse an empty guiUrl value (throws an exception)


fjoubert:
Like discussed, we only check the CA rights on katgui at the moment.

PR #289, opened 2017-03-15
Added subarray in maintenance section to the utilisation report when exporting as a pdf

fjoubert:
This section was overlooked when the report was developed.

screen shot 2017-03-15 at 7 20 41 pm


fjoubert:
Also in this PR: changed the start and end times to use UTC time, previously it was using SAST.

PR #292, opened 2017-03-16
Enable health views from config

fjoubert:
See https://github.com/ska-sa/katcamconfig/pull/189.

The PR above supplies the config needed for building health displays dynamically. In pictures, the config above gives us a dynamic list of health views in the left navigation bar:
screen shot 2017-03-16 at 3 55 46 pm
Clicking one of those links takes you to:
screen shot 2017-03-16 at 3 55 52 pm
The above display will display a list of defined tree views as defined in config, the same way the receptor display shows lists of treeviews.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -95,6 +95,8 @@
             read_only: 'Monitor Only'
         };
 
+        $rootScope.configHealthViews = [];
+
         SessionService.recoverLogin();
 
         $rootScope.getSystemConfig = function (forceConfig) {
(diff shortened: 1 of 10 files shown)


ajoubertza:
That's a nice addition.  You're not adding it to the home page, so what's the general approach regarding items available on the home page versus the menu?  Most seem to be in both places, but not all.  
(They're also in a different order/grouping, which is a bit confusing for me, but there's probably a reason for that.)

fjoubert:
It's not all thought out as well as it could be, but the basic reasoning is that the less used displays would go in the left navbar so that the home page is not too cluttered. I'll have to look at the order + grouping and try to cleanup the home page in the future - feel free to make a jira ticket!

PR #295, opened 2017-04-03
Value timestamp column for sensor list and misc.

fjoubert:
Minor pixel tweaks for scheduler toolbars also:

Configurable value timestamp on sensor list:
screen shot 2017-04-03 at 5 40 30 pm

Counter for resources assigned to a subarray:
screen shot 2017-04-03 at 4 47 43 pm
screen shot 2017-04-03 at 4 47 58 pm

### DIFF
--- a/app/scheduler/observations/observations-detail.html
+++ b/app/scheduler/observations/observations-detail.html
@@ -5,13 +5,14 @@
                 ng-class="{'md-hue-2': vm.subarray.state !== 'inactive'}" class="md-whiteframe-z2" layout="row"
                 layout-align="center center" title="{{vm.subarray.description}}" style="position: relative; height: 80px">
                 <div ng-include="'app/scheduler/templates/subarray-config-container.html'"></div>
-                <md-menu style="margin-top: -16px; margin-bottom: 0; padding: 0;">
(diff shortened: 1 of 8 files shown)


app/scheduler/observations/observations-detail.html
```diff
@@ -130,6 +131,9 @@
         </md-toolbar>
         <div flex class="schedule-block-resource-window md-whiteframe-z1" style="max-width: 450px; min-height: 284px; overflow: auto; min-width: 150px"
             layout="column" layout-wrap layout-align="start start">
+            <div class="resource-counter-overlay" title="The amount of resources assigned to this subarray">
+                {{vm.subarray.allocations.length}}
+            </div>
             <div ng-repeat="resource in vm.subarray.allocations track by $index" ng-class="{'maintenance-bg-hover': resource.maintenance}"
```
lvdheever: I know it is not so easy as this, but it will be more useful if we can show the nr of receptors, not the nr of resources. I was thinking that we could even add that as a sensor to subarray then it is also available in iPython.

fjoubert: Agreed that would be better. This was a quick and easy addition, so I think we should leave the ants count for a little later, maybe the busy week.

PR #296, opened 2017-04-19
States of resources on subarray activate and misc

fjoubert:
- listen to state sensors on resources
- display state of resources during activation with text and colour
- improved layout of gui-links display on subarray resources
- added load last known subarray configuration button for quick access
- increased the size and tweaked the subarray config template layout,
  also added opacity
- decreased size of resource items in subarray resources
- changed title of route Subarrays.Resource Assignment -> Subarrays.Resources
- changed title of route Subarrays.Observations Overview -> Subarrays.Overview
- adjusted min-width for subarray displays to avoid bad overlap of elements
- fix displaying of subarray config items on subarray overview
- change counter in subarray resources and observation detail to only
  count receptors assigned to subarray

Here's a flashy gif:
apr-19-2017 15-57-24

Here's a screenshot when they go into error state:
screen shot 2017-04-19 at 3 57 56 pm

### DIFF
--- a/app/scheduler/scheduler-home.js
+++ b/app/scheduler/scheduler-home.js
@@ -16,7 +16,7 @@
         UserLogService, NotifyService, MonitorService, ConfigService, $stateParams, $q, $mdDialog, UserService) {
 
         var vm = this;
-        var receptorRegex = new RegExp('^(m0..|ant.)$');
+        var receptorRegex = new RegExp('^(m|ant|dsh)\\d{1,4}$');
         vm.childStateShowing = $state.current.name !== 'scheduler';
         vm.subarrays = ObsSchedService.subarrays;
         vm.programBlocks = ObsSchedService.programBlocks;
@@ -828,9 +828,8 @@
                     classes += ' resource-state-activated';
                 } else if (resourceStateSensor.value === 'error') {
                     classes += ' resource-state-error';
-                } else if (resourceStateSensor.value === 'configuring' ||
-                           resourceStateSensor.value === 'configured' ||
-                           resourceStateSensor.value === 'activating') {
+                } else if (['deactivating', 'configuring', 'configured', 'activating']
+                           .indexOf(resourceStateSensor.value) > -1) {
                     classes += ' resource-state-busy';
                 }
             }
--- a/app/scheduler/scheduler-home.js
+++ b/app/scheduler/scheduler-home.js
@@ -16,6 +16,7 @@
         UserLogService, NotifyService, MonitorService, ConfigService, $stateParams, $q, $mdDialog, UserService) {
 
         var vm = this;
+        var receptorRegex = new RegExp('^(m0..|ant.)$');
         vm.childStateShowing = $state.current.name !== 'scheduler';
         vm.subarrays = ObsSchedService.subarrays;
         vm.programBlocks = ObsSchedService.programBlocks;
@@ -29,6 +30,7 @@
         vm.iAmCA = false;
         vm.modeTypes = ['queue', 'manual'];
         vm.guiUrls = ObsSchedService.guiUrls;
+        vm.resourcesStates = ObsSchedService.resourcesStates;
 
         if (!$stateParams.subarray_id) {
             $state.go($state.current.name, {
@@ -107,6 +109,18 @@
                 vm.subarray = null;
             }
         });
+
+        vm.getTotalReceptorsInSubarray = function () {
+            var counter = 0;
+            if (vm.subarray && vm.subarray.allocations) {
+                vm.subarray.allocations.forEach(function (resource) {
+                    if (receptorRegex.test(resource.name)) {
+                        counter++;
+                    }
+                });
+            }
+            return counter;
+        };
 
         vm.currentState = function() {
             return $state.current.name;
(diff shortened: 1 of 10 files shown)


app/scheduler/scheduler-home.js
```diff
@@ -16,6 +16,7 @@
         UserLogService, NotifyService, MonitorService, ConfigService, $stateParams, $q, $mdDialog, UserService) {
 
         var vm = this;
+        var receptorRegex = new RegExp('^(m0..|ant.)$');
```
ajoubertza: We could be a little stricter, using digits.  Could even allow for SKA dishes like `dsh0043`, if you like.
`^(m\d\d\d|ant\d|dsh\d\d\d\d)$`
(Although it isn't very pretty)

app/scheduler/scheduler-home.js
```diff
@@ -800,6 +814,29 @@
...
+            if (resourceStateSensor) {
+                if (resourceStateSensor.value === 'activated') {
+                    classes += ' resource-state-activated';
+                } else if (resourceStateSensor.value === 'error') {
+                    classes += ' resource-state-error';
+                } else if (resourceStateSensor.value === 'configuring' ||
+                           resourceStateSensor.value === 'configured' ||
+                           resourceStateSensor.value === 'activating') {
+                    classes += ' resource-state-busy';
+                }
```
ajoubertza: Are the "deconfigured" and "deconfiguring" states handled by default?  I see they are grey in the screen shot.

app/services/obs-sched-service.js
```diff
@@ -467,6 +467,9 @@
                         });
                     }
                 }
+            } else if (sensorName.endsWith('_state')) {
+                var component = sensorName.split('_state')[0];
+                api.resourcesStates[component] = sensor;
```
ajoubertza: Not sure what the possible sensor names are that could end up here.  If you are running through all sensors on a proxy, you could get something like `sdp_1_spmc_my_state`.  Would that end up in a component called `sdp_1_spmc_my`?

app/scheduler/scheduler-home.js
martinslabber: It would be good to apply the resource-state-busy style to deconfiguring.
It will help show who's taking long on free subarray.

app/scheduler/scheduler-home.js
```diff
@@ -108,6 +110,18 @@
             }
         });
 
+        vm.getTotalReceptorsInSubarray = function () {
```
martinslabber: See Jira CB-2055 I think this value should be supplied by subarray.

app/scheduler/scheduler-home.js
lvdheever: The "hard m0" is not nice and will bite is going forward. At least make this "m..."

app/scheduler/scheduler-home.js
lvdheever: One think I noticed in the very cute flashy gif is that the antennas show "activated" when the subarray is in "idle" . This feels wrong. It may be something that has to be corrected in the receptor proxy, or subarray to send "deactivate" to receptor (I have not looked at how the receptor state has been implemented)

app/services/obs-sched-service.js
lvdheever: It is possible, but the way it is "extracted" from api.resourcesState is by indexing with a component name. So it is possibly acceptable to not do anything more complicated here. May be just a comment noting that this will store more than is necessary, for the enquiring minds of other software engineers?

ajoubertza:
I wonder what will happen with the old data proxy we use on RTS.  It doesn't have a `state` sensor.

app/scheduler/scheduler-home.js
fjoubert: deconfigured and deconfiguring applies no colours to the resources - they are transparent by default. I will add deconfiguring to the purple colour. FYI I got these directly from the document https://docs.google.com/document/d/1Xub60S_MdiIbLxHWSoAdIHFlRvSufS8uL6CERD8Z2ns/edit#bookmark=id.e2curux1b2ou , so maybe that needs updating too.

app/scheduler/scheduler-home.js
fjoubert: Good points, will change it 👍

app/services/obs-sched-service.js
fjoubert: Yes, only actual resources will get the state from the dictionary. The idea was that the regex that katportal uses when setting up the strategies and subscribing should be enough to filter out the other sensors so that the gui does not need to know about here. This works because we use a specific namespace for the _state sensors and some subarray sensors.

app/scheduler/scheduler-home.js
fjoubert: Agreed, until then we have this :)

fjoubert:
Any resources that doesn't have a state will simply stay transparent and not display a state - like `anc` does now.

app/scheduler/scheduler-home.js
ajoubertza: Doc updated.

The receptor proxy state sensor was added recently, with a trivial implementation to get things started: it is always "activated".  We haven't implemented the state machine there yet.  Hope to improve it in soon.

lvdheever:
I just wondered the same thing about "state" for RTS on the drive here. Good to hear that it will be OK.

app/scheduler/scheduler-home.js
martinslabber: @lvdheever the receptor state sensor is hard coded at the moment see CB-2022.
We did that in sprint planning to not hold up the GUI work - https://github.com/ska-sa/katproxy/pull/351

lvdheever: OK - no problem then. Please make a backlog for improving receptor status sensor if there isn't one yet.

PR #301, opened 2017-07-31
Added url query params for sensor list for shareable links

fjoubert:
Also fixed a layout issue with the channel-mask sensor, where the popup dialog got squashed.

screen shot 2017-07-31 at 11 01 42 am

### DIFF
--- a/app/sensor-list/sensor-list.js
+++ b/app/sensor-list/sensor-list.js
@@ -3,24 +3,26 @@
     angular.module('katGui')
         .controller('SensorListCtrl', SensorListCtrl);
 
-    function SensorListCtrl($scope, $rootScope, SensorsService, $timeout, KatGuiUtil, $interval,
-                            $log, $mdDialog, DATETIME_FORMAT, NotifyService, ConfigService, $localStorage) {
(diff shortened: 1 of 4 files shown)


app/sensor-list/sensor-list.js
```diff
@@ -3,24 +3,26 @@
...
         vm.connectInterval = null;
+        vm.initDone = false;
 
         vm.showTips = false;
         vm.showContextZoom = false;
         vm.useFixedYAxis = false;
         vm.yAxisMinValue = 0;
         vm.yAxisMaxValue = 100;
-        vm.hideNominalSensors = false;
+        vm.hideNominalSensors = $stateParams.hideNominal? $stateParams.hideNominal === 'true': false;
```
martinslabber: This seems a bit dodgy but probably makes no difference. 
Sometimes you return a string and other times a bool.
Either the eval condition should lose the "'" or the else result should be wrapped.

fjoubert: This is a null check for the input param, this makes it optional to use. All the input parameters from the query string is passed in as strings, so we need to check for that specifically with `$stateParams.hideNominal === 'true'`, which returns a bool, maybe the syntax is strange that you think it returns a string, because it only ever returns a bool.

PR #312, opened 2017-09-20
Added a tags query parameter to allow for tag filtering by name instead of just tag.id

fjoubert:
By request of Rupert.

### DIFF
--- a/app/userlogs/userlog-reports.js
+++ b/app/userlogs/userlog-reports.js
@@ -172,6 +172,11 @@
                         vm.filterTags = UserLogService.tags.filter(function (tag) {
                             return tagIdsList.indexOf(tag.id.toString()) > -1;
                         });
+                    } else if ($stateParams.tags) {
+                        var tagsList = $stateParams.tags.split(',');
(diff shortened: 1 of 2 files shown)


app/userlogs/userlog-reports.js
```diff
@@ -172,6 +172,11 @@
                         vm.filterTags = UserLogService.tags.filter(function (tag) {
                             return tagIdsList.indexOf(tag.id.toString()) > -1;
                         });
+                    } else if ($stateParams.tags) {
+                        var tagsList = $stateParams.tags.split(',');
+                        vm.filterTags = UserLogService.tags.filter(function (tag) {
+                            return tagsList.indexOf(tag.name) > -1;
+                        });
```
lvdheever: If you specifies tag id's it ignores any tag names, if I read this correctly. Is that the intention and desired result?

fjoubert: Yes, I didn't want to mix them. So, if you specify tag ids I assume you know what you are doing and the tags query parameter is ignored.

PR #319, opened 2018-01-02
Reworked reacting to lead operator changes to be inline with reworked katportal behaviour

fjoubert:
Also fixed a reconnection bug not updating connection status properly because of a missing promise.

Related katportal PR https://github.com/ska-sa/katportal/pull/298


fjoubert:
LO tests look good - going to do some additional final testing before this and the katportal branches are merged.

PR #322, opened 2018-01-09
MKAT - Change list-sensors websocket rpc calls to be HTTP GET's instead

fjoubert:
Also in this PR:

- Fix CAM Components display process start,stop,restart,kill actions
- Fixes for tooltips on hover for the receptor health displays
- Removed some old unused health displays
- Reworked the layout of the process-control display
This PR depends on the katportal PR - https://github.com/ska-sa/katportal/pull/302


ajoubertza:
This is a very large set of changes to hotfix.  What testing has been done to verify it works?
- Tested on a multinode system to verify the nodeman process control changes? 
- Changes to scheduler - run observations?
- Changes to operator controls - tested stow, shutdown?
etc.

fjoubert:
@ajoubertza extensive testing for each affected display was done on dev64, you can try the katportal and katgui branches there too if you have some time.

ajoubertza:
When I activate a subarray on dev64 (cbf, sdp and 3 antennas) the GUI shows a timeout message.  Subarray still activates successfully, but would be nice not to get the error unnecessarily.
image
Not sure if this is a GUI, portal or subarray issue.

ajoubertza:
On the receptor health page, if I click on a sunburst the text goes away.  Clicking doesn't make it come back.  It that a feature?
image

ajoubertza:
On the process control page there seem to be many processes missing, e.g. CMC, SPMC, TFR. Nothing under the SIM node (possibly where these run).  Also notice the node names are sorted in reverse order.  Would prefer alphabetical.  I'm used to monctl being first. It is fine if the proxy nodes come later, but would also prefer them in ascending order.
image

ajoubertza:
I stopped the M010 receptor proxy process and restarted it again via the GUI.  It restarted, but after that all of its RSC sensors were unreachable.  The connection from the M010 proxy to its RSC simulator is reported is disconnected. Probably not a GUI issue though!  On my single node system, a restart like this works fine.

fjoubert:
- Subarray activation error is a katcp request timeout between portal and subarray_N
- Missing text is clearly a bug, will check it out
- I'll play a bit more with the order of the process control, it sorts the proxies last then by the number of processes in the NM. The nm_sim not showing processes is a problem with nm_sim sensors (going to do a restart to fix)
- Receptor proxies not restarting properly is usually because of the RSC simulator not resyncing properly.

Only gui fix from the above is the missing text, but I'll do a stop/start to see the effect that has on the other issues.

ajoubertza:
OK, just the text to be fixed for now.

fjoubert:
Fixed the text in the sunburst and reordered the process control display, see screenshots:

screen shot 2018-01-10 at 10 42 31
screen shot 2018-01-10 at 10 43 15

PR #325, opened 2018-01-12
Allow dump rate to be set from the gui

fjoubert:
Related JIRA https://skaafrica.atlassian.net/browse/CB-2056

Related PR's:
- https://github.com/ska-sa/katportal/pull/303
- https://github.com/ska-sa/katconfig/pull/622

Added a dropdown to allow the selection of the dump rate of the user product from the gui
screen shot 2018-01-12 at 10 41 34
screen shot 2018-01-12 at 11 19 44
screen shot 2018-01-12 at 10 47 56

- The default value is what the subarray specifies (0.25hz) and is reset to that when the subarray is freed.
- The dump rate is shown in hz with seconds in parentheses next to it to avoid hz -> seconds conversion mistakes and indicates which of the values is the default (as per config)
- Allowable dump rates is config driven and should be changed at some point. For now a list of defaults has been assumed to be one of 0.125, 0.25, 0.5, 0.75, 1.0 for each user product (see https://github.com/ska-sa/katconfig/pull/622)
- When changing the user product, the dump rate is set to its specified default in katconfig
- When changing the user product and the current selected band is not in the list of allowable bands for the new user product, the band will default to the first allowable band in the new user product.
- Changed the text of 'Config Label' to just be 'CL' - there's a tooltip on that element reads 'Config Label'


app/services/obs-sched-service.js
```diff
@@ -410,7 +413,8 @@
                                                 return resource[0];
                                             }).join(","),
                                         band: api.subarrays[subarrayIndex].band,
-                                        product: api.subarrays[subarrayIndex].product
+                                        product: api.subarrays[subarrayIndex].product,
+                                        dump_rate: api.subarrays[subarrayIndex].dump_rate
                                     };
```
lvdheever: Not sure where the order of the subarray title bar is determined, but the screen shot shows "product, band, dump_rate", and it should rather be "band, product, dump_rate"

app/scheduler/scheduler-home.js
```diff
@@ -88,13 +91,35 @@
...
                 var productKeys = Object.keys(productConfig);
                 productKeys.forEach(function(product) {
                     vm.products.push({
                         name: product,
                         sp_product: productConfig[product].sp_product,
                         cbf_product: productConfig[product].cbf_product
                     });
+                    if (productConfig[product].allowed_dumprates) {
+                        vm.dumpRatesMap[product] = productConfig[product].allowed_dumprates.split(',').map(
+                            function(dumpRate) {
```
lvdheever: Did you make sure to initialse the drop down with the **default** dump_rate to ensure that the user gets the default when he does not set/select it.

app/services/obs-sched-service.js
fjoubert: The order in this function does not matter, it's a dict that is saved to localStorage for the 'Load last known configuration' button.

But for the gui, I changed the order to be product, band, dump_rate because of the limitations that a product has regarding bands and dump_rates. In katconfig we have an 'allowed_bands' field, so I limited the bands in the dropdown to only list the allowable bands for each product and to force you to select a product if you select a band.

So it made sense to me that having the product on the left would guide a user in selecting the product before the band and before the dump rate. There is still a work around for this by clicking on the context menu and using the 'Select Band...' dialog.

No product selected:
screen shot 2018-01-12 at 11 57 29

Product selected with allowable bands:
screen shot 2018-01-12 at 11 57 37
screen shot 2018-01-12 at 11 57 45

Workaround for selecting a band in any case:
screen shot 2018-01-12 at 12 01 06
screen shot 2018-01-12 at 12 00 58

I can remove the part about only being able to select allowable bands if we do not want to limit the band selection like this.

app/scheduler/scheduler-home.js
fjoubert: The default only comes into play when you change a product. So if you have a dump rate set to 1hz and change the product, the dump rate is then set to the default of the new product.

If there are no allowed_dumprates in the config, the dropdown is initialised with the default dumprate for the selected product as the only item in the list. If no product it selected it shows a text item asking for a product to be selected:
screen shot 2018-01-12 at 12 09 21

When you free a subarray the dump rate is reset by the subarray (it seems the default is always 0.25 when the subarray is started and freed again - I didn't double check that)

fjoubert:
@lvdheever if you want to try it out, feel free to click around on http://monctl.devf.camlab.kat.ac.za/katgui/scheduler/resources/1

app/services/obs-sched-service.js
lvdheever: I am a little worried about this restriction of user_product to band. I don't think that is how it was, but I may be mistaken -  user could select a band first, then any product (I am not sure that our config is complete with regards the allowed bands defined for each user product). Will have to check with @lmagnus first before we push this restriction out to site.
If allowed_bands in the user_product configs are correct, then what you did here is fine.
But still would like to see the product and dump_rate next to each other - so may be the order should then be "product, dump_rate, band" instead of "product,band, dump_rate".
But let's chat with Lindsay about they want to use it - select product, then the rest, as you have it now; or the other way round: select band, then restrict the products to the selected band.

fjoubert: Sounds good to get input from the users.

From what I can see in https://github.com/ska-sa/katconfig/blob/karoo/user/products/karoo.mkat.conf it looks like all products has an allowed_bands item.

There's no rush to merge this in today.

PR #329, opened 2018-01-19
Added pb deletion (with orphaning sb's option), pb assign/unassign to program block display.

fjoubert:
Also:
- Added PB time estimate to the observation schedule pb's headers
- Replaced sometimes buggy md-virtual-repeats with plain ng-repeats


app/services/obs-sched-service.js
```diff
@@ -309,6 +330,29 @@
...
+                    var momentDuration = moment.duration(pb.expected_duration_seconds, 'seconds');
+                    var durationHours = momentDuration.hours().toString().padStart(2, 0);
+                    var durationMinutes = momentDuration.minutes().toString().padStart(2, 0);
+                    var durationSeconds = momentDuration.seconds().toString().padStart(2, 0);
+                    var durationStr = durationHours + ':' + durationMinutes + ':' + durationSeconds;
+                    var durationDays = momentDuration.days();
+                    if (durationDays) {
+                        durationStr = durationDays + 'd ' + durationStr;
+                    }
+                    pb.expectedDuration = durationStr;
```
ajoubertza: One all the SBs are done, will this expectedDuration get cleared or remain at the total time?

fjoubert: It gets recalculated every time the observation schedule changes, so it will return to 0. That doesn't really matter though, because a PB is no longer shown when none of it's SB's are in the observation schedule.

app/services/obs-sched-service.js
```diff
@@ -309,6 +330,29 @@
...
+                        pb.expected_duration_seconds += sb.expected_duration_seconds;
+                    }
+                });
+                if (pb.expected_duration_seconds) {
+                    var momentDuration = moment.duration(pb.expected_duration_seconds, 'seconds');
+                    var durationHours = momentDuration.hours().toString().padStart(2, 0);
+                    var durationMinutes = momentDuration.minutes().toString().padStart(2, 0);
+                    var durationSeconds = momentDuration.seconds().toString().padStart(2, 0);
+                    var durationStr = durationHours + ':' + durationMinutes + ':' + durationSeconds;
+                    var durationDays = momentDuration.days();
```
lvdheever: I suspect for duration hours only may be better (not converting it to days) even if it takes more than 24hrs, but lets check with commissioners/operations. OK to leave like this for now, can update at another cycle if necessary.

@lmagnus - your thoughts?  Total ProgramBlock duration - better to display in hours:min:seconds or days:hours:min:seconds?

app/services/obs-sched-service.js
```diff
@@ -122,6 +122,26 @@
             });
         };
 
+        api.deleteProgramBlock = function(pb, orphanSBs) {
+            var orphanDialogText = orphanSBs? " and orphan it's uncompleted Schedule Blocks?" : " and all it's uncompleted Schedule Blocks?";
+            NotifyService.showImportantConfirmDialog(
```
lvdheever: We cannot allow deletion of a PB that has some completed SBs. We **can** have an option to orphan SBs from a PB, but once you used a PB to observe some SBs, that PB is untouchable.

fjoubert: Ok I'll change that on katportal.

fjoubert: See https://github.com/ska-sa/katportal/pull/313 for the fix.

PR #331, opened 2018-01-22
MKAT - Fix broken schedule block approved display

fjoubert:
Also fixed broken warn log and added 'track by' to some ng-repeats.


fjoubert:
Creating another PR with some more fixes in.

PR #332, opened 2018-01-22
Fixed some overflow scrolling in scheduler displays

fjoubert:
Also:
- Display PB expected duration only as hh:mm:ss
- built new version


lvdheever:
I prefer it if the master and release/branch PRs that implement the same things are named similarly so that it is easier to confirm that a release/branch changes are not "lost" in master.

fjoubert:
Yeah my bad - I accidentally pushed some of the changes to master earlier without making a branch, that is why the two PR's look so different.

PR #339, opened 2018-02-08
Fix the ganglia navigation button and enable the video button correctly

fjoubert:
- The ganglia navigation button lost its `ng-click` handler sometime in the past. I've added it back in.
- The video button was being disabled by checking old config values, which was changed when we started supporting more than 1 video feed.
- Fixed a bug in the sensor-groups display that sometimes does not show the sensor name if the original katcp name is missing from the sensor cache (which is fixed in katcore master).

Jira - https://skaafrica.atlassian.net/browse/CB-2699


lvdheever:
Tx for the quick response

PR #346, opened 2018-02-15
Adding duration of subarray in maintenance on the pdf report

bxaia:
@fjoubert please review

This pull request addresses adding duration of the subarray set to maintenance on the PDF generated for utilisation report.


app/reports/utilisation-report.js
```diff
@@ -114,10 +114,10 @@
...
                 rows = [{
                     value: 'In Maintenance',
-                    percentageOfTotal_1: vm.subarrayMaintenanceDurations['1'].percentageOfTotal || '',
-                    percentageOfTotal_2: vm.subarrayMaintenanceDurations['2'].percentageOfTotal || '',
-                    percentageOfTotal_3: vm.subarrayMaintenanceDurations['3'].percentageOfTotal || '',
-                    percentageOfTotal_4: vm.subarrayMaintenanceDurations['4'].percentageOfTotal || ''
+                    percentageOfTotal_1: (vm.subarrayMaintenanceDurations['1'].percentageOfTotal || '') + '(' + vm.subarrayMaintenanceDurations['1'].duration')',
+                    percentageOfTotal_2: (vm.subarrayMaintenanceDurations['2'].percentageOfTotal || '') + '(' + vm.subarrayMaintenanceDurations['2'].duration')',
+                    percentageOfTotal_3: (vm.subarrayMaintenanceDurations['3'].percentageOfTotal || '') + '(' + vm.subarrayMaintenanceDurations['3'].duration')',
+                    percentageOfTotal_4: (vm.subarrayMaintenanceDurations['4'].percentageOfTotal || '') + '(' + vm.subarrayMaintenanceDurations['4'].duration')'
```
ajoubertza: @lvdheever Not related to this PR, but this part of the code won't extend to more subarrays automatically.

fjoubert: There's a couple of places in this pdf report that will only show the 1st 4 subarrays - should be pretty straight forward to use config that the gui already has - I'll see if I can quickly add it from config instead of this hard coding.

fjoubert: The syntax here is broken. There's a missing `+`. It should be
```
percentageOfTotal_4: (vm.subarrayMaintenanceDurations['4'].percentageOfTotal || '') + '(' + vm.subarrayMaintenanceDurations['4'].duration + ')'
```

I'll fix it up now when I get the subarray numbers from config.

fjoubert:
f9e39ff should fix the hardcoded subarrays - @bxaia is going to verify that it is working as intended (worked on my machine).

app/reports/utilisation-report.js
lvdheever: Yes, I noticed that earlier and have a note about it, but may be Francois fixes it now to get subarrays from config

fjoubert:
@bxaia checked and confirmed that it works as intended

PR #349, opened 2018-02-28
User/ofaletse/get vds constants from from portal

ofaletse:
@fjoubert please review.

This PR extract  vds constants, and bind the max value of focus and zoom slider. This is so to avoid hard coding the zoom and focus max values.
This is per request from Caxton in the email.


app/video/video.js
```diff
@@ -14,6 +14,7 @@
         vm.imageSource = null;
         vm.sensorValues = {};
         vm.stepTimeValue = 1;
+        vm.vdsConstants = {};
```
fjoubert: This is fine, but duplicates the constants object. You could assign it to the object in the service instead like `vm.vdsConstants = ConfigService.vdsConstants`, then when you load the vds constants data, you don't need to assign anything - i.e. line 39-40 could simply be `ConfigService.loadVDSConst();`.

The biggest reason why you would want to use the ConfigService variable instead is so that you can check if the VDS constants was loaded before, then you can just return that in the `ConfigService.loadVDSConst()` call instead of making another GET request. See the comment in the ConfigService class.

app/services/config-service.js
```diff
@@ -34,6 +35,19 @@
...
+            $http(createRequest('get', urlBase() + '/vds-config'))
+                .then(function (result) {
+                    api.vdsConstants = result.data;
+                    deferred.resolve(api.vdsConstants);
+                }, function (message) {
+                    $log.error(message);
+                    deferred.reject();
+                });
+            return deferred.promise;
+        };
```
fjoubert: You could change this to only ever call the config service once if you do the same thing `api.loadAggregateSensorDetail` does:
```
        api.vdsConstants = null;

        ...

        api.loadVDSConst = function () {
            var deferred = $q.defer();
            if (api.vdsConstants) {
                deferred.resolve(api.vdsConstants);
            } else {
                $http(createRequest('get', urlBase() + '/vds-config'))
                    .then(function (result) {
                        api.vdsConstants = result.data;
                        deferred.resolve(api.vdsConstants);
                    }, function (message) {
                        $log.error(message);
                        deferred.reject();
                    });
            }
            return deferred.promise;
        };
```

Just make sure that this still works as expected when the object is initialised as null (`api.vdsConstants = null;`) and then the reference to the object is used in the VDS controller function.

app/video/video.js
lvdheever: @ofaletse Do you still need to make this change `vm.vdsConstants = ConfigService.vdsConstants`?

lvdheever:
@ofaletse Don't you need changes in video.html to now use these constants for pan, tilt, focus and zoom: like: max="{{vm.vdsConstants[vm.vds_name].moxa_max_focus}}"
Or is that coming in a separate PR?

And it looks like the assumption is that these min/max values for pan, tilt, focus and zoom **has** to be present in the config file as there are now defaults loaded in loadVDSConstants and no safety built into the html where it is used. I suggest then to add a comment to the config file on which fields are mandatory as it looks quite "benign" and as if it is not that important when you see the config file by itself.

ofaletse:
@lvdheever `pan` and `tilt` are not using sliders, so we do not need to bind the constants.
Please see [commit](https://github.com/ska-sa/katcamconfig/commit/7a218df)

app/video/video.html
```diff
@@ -42,12 +42,12 @@
                     </div>
                     <div layout="row" layout-align="start center" style="margin-top: 8px; width: 268px;">
                         <span style="margin-right: 6px; width: 100px">Focus:</span>
-                        <md-slider class="md-primary" style="background: transparent; width: 150px; margin-left: 8px;" min="0" max="40" ng-model="vm.focus" ng-change="vm.focusChanged(vm.focus)">
+                        <md-slider class="md-primary" style="background: transparent; width: 150px; margin-left: 8px;" min="0" max="{{vm.vdsConstants[vm.vds_name].moxa_max_focus}}" ng-model="vm.focus" ng-change="vm.focusChanged(vm.focus)">
```
ajoubertza: The config also has minimum values for focus and zoom, so why don't we use those?  Here, and line 50.

ofaletse:
@ajoubertza I do not use the minimum values in the video gui for readability(for the sake of the user).
This is handled in the vds translator. Please review when I push.

app/video/video.html
lvdheever: However, the slider will be more accurate if you also use the min value here, and since you already use the max value from config I see no reason why not to also use the min value. The min value is sometimes quite substantial (like moxa_min_zoom: 59 with moxa_max_zoom: 957, almost 10%) and will make a visible difference in how the slider displays.

PR #355, opened 2018-03-16
Merge sensor-list updates

lvdheever:
This brings in the camv19 changes of sensor-list to master.
HideWarn checkbox and only filter on Search box on submit (pressing enter)

(@ajoubertza please note as you are fiddling with sensor-list)


app/sensor-list/sensor-list.html
```diff
@@ -1,3 +1,5 @@
+ <form novalidate name="frm1" autocomplete="off" ng-submit="vm.searchFilter=searchText">
```
ajoubertza: As we noticed, this affects the scrolling of the layout - components, proxy and sensor list header are no longer static.  Would be good to fix this.

app/sensor-list/sensor-list.js
```diff
@@ -273,16 +275,21 @@
...
+            return show
         };
 
         vm.updateURL = function () {
-            $state.go('sensor-list', {
-                component: vm.resourceSensorsBeingDisplayed? vm.resourceSensorsBeingDisplayed: null,
-                filter: vm.searchFilter? vm.searchFilter: null,
-                hideNominal: vm.hideNominalSensors? 'true': null},
-                { notify: false, reload: false });
+            if (!vm.noSensorUpdates) {
```
ajoubertza: What causes this variable to change?

app/sensor-list/sensor-list.html
lvdheever: Yes, I added this to the list of KatGUI fixes.

app/sensor-list/sensor-list.js
lvdheever: Oops - this is a remnant of a previous attempt to have a checkbox to start/stop sensor udpates, which proved to be more cumbersome than the "Enter" approach. Will remove.

PR #359, opened 2018-03-18
Fixed build errors and improved some logic

fjoubert:
- Fixed various syntax errors
- Improved config-health-view.js logic for finding the diff of 2 sensor lists.
- Built a new version

This after a request from @lvdheever to help out with some build errors. The errors were caused by syntax errors. Mainly in the sensor-list.js file (see https://github.com/ska-sa/katgui/compare/user/fjoubert/fix-syntax-and-build?expand=1#diff-704cc227e989581ee58519d848cbf415R314).

There were some other concerning bits that I had a look at that hopefully helps.
Unfortunately my linter got hold of a couple of files so this PR is larger than it needs to be 😢


lvdheever:
Thanks @fjoubert!  I did not expect you to do a code review, but much appreciated. I've tested this and will merge to fix up master.

PR #360, opened 2018-03-19
User/ofaletse/avoid restricting focus slider max value

ofaletse:
@lvdheever please review

This PR avoid restricting focus slider max value(40), and changed it to use 900.
NB I will close [349](https://github.com/ska-sa/katgui/pull/349)  and [326](https://github.com/ska-sa/katportal/pull/326)


app/video/video.html
```diff
@@ -42,7 +42,7 @@
                     </div>
                     <div layout="row" layout-align="start center" style="margin-top: 8px; width: 268px;">
                         <span style="margin-right: 6px; width: 100px">Focus:</span>
-                        <md-slider class="md-primary" style="background: transparent; width: 150px; margin-left: 8px;" min="0" max="40" ng-model="vm.focus" ng-change="vm.focusChanged(vm.focus)">
+                        <md-slider class="md-primary" style="background: transparent; width: 150px; margin-left: 8px;" min="0" max="900" ng-model="vm.focus" ng-change="vm.focusChanged(vm.focus)">
```
lvdheever: These seem like "arbitrary" choices. The GUI should reflect the config values which you've added recently. Suggest to load it into the vm e.g. as vm.configvalues[vm.vds_name].zoom_min, zoom_max, focus_min and focus_max or something similar and populate the HTML from the config values.

package-lock.json
```diff
@@ -0,0 +1,6750 @@
+{
```
lvdheever: We've never added this package-lock file before. Is it necessary, or did it just slip in?

app/video/video.html
ofaletse: Sure, you are right, will change that.

package-lock.json
ofaletse: It slipped in, but I really wonder what is it for. #wondering
I will remove it, and furthermore try to understand what is it for.

app/video/video.js
```diff
@@ -15,6 +15,14 @@
...
         vm.stepTimeValue = 1;
         vm.sensorsRegex = '^anc_vds';
+        vm.focusParams = {
+          'min': 0,
+          'max': 900
+        }
+        vm.zoomParams = {
+          'min': 0,
+          'max': 905
+        }
```
lvdheever: In the previous PR #349 you used `max="{{vm.vdsConstants[vm.vds_name].moxa_max_focus}}"` which is what it should be - use the katcamconfig values you added recently instead of arbitrary hardcoded values.

ofaletse: Thanks for taking your time and review, I hear you, but I re-applied my mind and do not think this is the direction that we should be heading towards. 
I will come to your desk to elaborate to you my point of view.

lvdheever: Discussed - the proxy does the scaling from the min/max values to the actual MOXA values. A comment here to reference where this is done and that these min/max values must be aligned with the proxy implementation and where it is used will suffice for now.

PR #361, opened 2018-03-22
handle subscribtion of dynamic sensor better

ofaletse:
@lvdheever or @ajoubertza please review.
This PR handle subscription of dynamic sensor(cbf) better and update the center button of the sunburst


app/health/config-health-view/config-health-view.js
```diff
@@ -126,7 +141,13 @@
...
                 return;
             }
-            MonitorService.listSensorsHttp('cbf_1,cbf_2,cbf_3,cbf_4', 'device_status', true).then(function (result) {
+            vm.oldTime = (new Date).getTime();
+
+            if (!vm.component) {
+                $log.info("Failed to grep component name");
+                return;
+            }
+            MonitorService.listSensorsHttp(vm.component, vm.cbfSensorRegx, true).then(function (result) {
```
lvdheever: Here - if you use "all" and not vm.component (like it is done in initSensors) then you don't need to derive a vm.component; e.g. `MonitorService.listSensorsHttp('all', view.sensors.join('|'))` where view is as per initSensors. This will query the MonitorService for the current list of sensors using the **full** sensor names as initialised in populateSensorNames (which now ensures that the full sensor name with the component prefix is used always - but did not do that originally).

I.e `MonitorService.listSensorsHttp(vm.component, vm.cbfSensorRegx, true)` becomes
```view = StatusService.configHealthSensors[vm.selectedConfigView];
MonitorService.listSensorsHttp('all', view.sensors.join('|')) ...
````
and then this should already include what you have defined in vm.cbfSensorRegx as those are included in the config health view configuration and thus are present in view.sensors.

File: app/health/config-health-view/config-health-view.js [hunk 2]
```diff
@@ -148,9 +168,7 @@
                             d3.selectAll('.' + sensor.name).attr('class', 'unknown' + '-child ' + sensor.name);
                         });
                     }
-                    vm.subscribedSensors.forEach(function (sensor) {
-                        MonitorService.unsubscribeSensor(sensor);
-                    });
+
```
lvdheever: Using just the length of the lists and assuming that the array with the longer length is adding sensors is a bit crude. I wonder if we can rather check only for new sensors in newList and subscribe to the new ones here, and let all the unsubscribe be handled by the destroy. 

Then the section below where it subscribes again to each sensor `for( var sensor in vm.newList) {` is unnecessary.

app/health/config-health-view/config-health-view.js
```diff
@@ -152,12 +153,7 @@
                             d3.selectAll('.' + sensor.name).attr('class', 'unknown' + '-child ' + sensor.name);
                         });
                     }
-
                     vm.subscribedSensors = [];
-                    for( var sensor in vm.newList) {
```
lvdheever: This "emtpying" of vm.subscribedSensors here should also be removed. See if you can pop the sensor off vm.subscribedSensors where you unsubscribe from it earlier.

PR #363, opened 2018-03-26
User/lize/sensor list updates

lvdheever:
Remove unused healthsub and healthcbf.
Use filter on enter through ng-keypress and setFilterOnEnter instead of making it a form and using submit - this re-instates the earlier scroll behaviour.


ofaletse:
Looks ok. Thanks for the change.

PR #365, opened 2018-03-29
Avoid unsubscribing some of the sensors when re-activating a subarray

ofaletse:
@ajoubertza please re-review

This is a fix that avoid unsubscribing sensors after re-activating.


ofaletse:
Yes, i normally wait for the a go-ahead from a reviewer. before I commit the dist folder code.

PR #370, opened 2018-04-18
Displaying icons on the legend

bxaia:
Added background colour on the icons for the legends.


ajoubertza:
@bxaia Can you please include a screenshot in the description showing the bit that has changed?

bxaia:
screen shot 2018-04-19 at 15 05 37

ajoubertza:
In the screenshot some of the receptors have white backgrounds.  Is that normal, or could it be related to the changes in this PR?  There are no white entries in the legend, which is why I ask.

lvdheever:
I've never seen white backgrounds - may be related to the changes.

bxaia:
screen shot 2018-04-20 at 12 56 05

I tried debugging the white space on the aggregate sensors for the time sync on Top by reverting my changes to the legend individually so I could see which one introduced the bug. But even after removing all of the changes I've made on the legend we still have the white background.
The sensors displaying white background are supposed to be unreachable which is a pinkish colour. 

I will log in a ticket for this bug as Ofaletse thinks it's something that would need extensive debugging as these are aggregate sensors.

lvdheever:
Please also check the other health views (Receptor, Custom & Config Health) if not yet done. Not sure if that detail made it into the split out "Legend" ticket.

bxaia:
@lvdheever I've checked all the other Health views, the legend changes are implemented there as well. However the white background is only on the Top Health

PR #371, opened 2018-04-18
Adding dist files for the commit on the kibana url time change

bxaia:
Adding dist files


ofaletse:
Yes you are right @lvdheever , I failed to pick that up. B may you please ```git add``` and ```git commit``` the two newly created untracked files in the dist folder.

PR #372, opened 2018-04-19
User/ofaletse/cb 2738/fix the drop down to only list program name

ofaletse:
@lvdheever please review.

This PR set the list of drop down to only contain program names, instead of listing all the files in `/var/kat/log`, which makes it very hard to navigate through the list of 21 files per process/program_name

image

[Jira](https://skaafrica.atlassian.net/browse/CB-2738)

As for dist files I will commit the files after Bulelalni's PR's are merged.


app/userlogs/userlog-reports.js
```diff
@@ -3,16 +3,21 @@
...
             vm.startTime = new Date();
             vm.endTime = vm.startTime;
             vm.startDatetimeReadable = moment(vm.startTime.getTime()).format('YYYY-MM-DD 00:00:00');
             vm.endDatetimeReadable = moment(vm.endTime.getTime()).format('YYYY-MM-DD 23:59:59');
             vm.tags = UserLogService.tags;
-            vm.logFiles = UserLogService.logFiles;
+            vm.logFiles = ['activity', 'alarm'];
+            vm.nodeNames = [];
+            vm.processSensors = ['running', 'args', 'pid', 'return_code'];
+            vm.sensorsRegex = vm.processSensors.join('|');
```
lvdheever: I am not sure why the subscription is to all the "running" "args" "pid" and "return_code" sensors for each processes. The nodeman makes that set of sensors for each of the processes that is registered and subscribing only the "running" is sufficient to get the complete list of processes, i.e. `vm.sensorsReges = ["running"]`

app/userlogs/userlog-reports.js
```diff
@@ -37,6 +42,30 @@
...
+            vm.getLogFiles = function () {
+                ConfigService.getSystemConfig()
+                    .then(function() {
+                        var nodeNames = Object.keys(ConfigService.systemConfig.nodes).map(function(node) {
+                            return 'nm_' + node;
+                        });
+                        MonitorService.listSensorsHttp(nodeNames.join(','), vm.sensorsRegex).then(function(result) {
+                            result.data.forEach(function(sensor) {
+                                MonitorService.subscribeSensor(sensor);
+                                if (sensor.name.endsWith('running') && sensor.original_name) {
```
lvdheever: Here you also only operate on the "running" sensors. So it seems like subscribing to the others is not required.

app/userlogs/userlog-reports.js
```diff
@@ -37,6 +42,30 @@
                 vm.endDatetimeReadable = moment(vm.endTime.getTime()).format(MOMENT_DATETIME_FORMAT);
             };
 
+            vm.getLogFiles = function () {
```
lvdheever: I have one concern because of my lack of knowledge about when getLogFiles will be called. If it is called too early and the nodeman processes are still being started up, then it will not be fully populated. It is in the body of the main "UserlogReportsCtrl" in this js file, but does that mean it gets initialised when you open the user log screen, or when the app starts. If it is the latter then it is not sufficiently "late". Also because the webservers have been moved to startup early, in the "services" group.
Please test by going to the userlog screen in the first 20-30 seconds after the kat-start when the system is still starting up.

app/userlogs/userlog-reports.js
```diff
@@ -37,6 +42,30 @@
...
+                            return 'nm_' + node;
+                        });
+                        MonitorService.listSensorsHttp(nodeNames.join(','), vm.sensorsRegex).then(function(result) {
+                            result.data.forEach(function(sensor) {
+                                MonitorService.subscribeSensor(sensor);
+                                if (sensor.name.endsWith('running') && sensor.original_name) {
+                                    // e.g. nm_monctl.anc.running
+                                    var splitName = sensor.original_name.split('.');
+                                    var processName = splitName[1];
+                                    vm.logFiles.push('kat.'+processName);
```
ajoubertza: We mentioned the webservers need to be updated to fix the program names.  Another special case is some of the katstore processes.  E.g. We have a sensor `nm_store.katstore_hdf.running`, but the program name for logging is `kat.katstore.hdfserver`.

app/userlogs/userlog-reports.js
ofaletse: @lvdheever Thanks for picking that up, yes that was pointless. will fix

app/userlogs/userlog-reports.js
ofaletse: Thanks for the comment, although there is a possibility for this to occur, I am of the view that it is unlikely for this to occur. 
My view is based on the fact that you can not use katgui when the system is starting up.

In answering the other question: it is initialized when one open the user log screen.

I did the test and all seem to be working.

app/userlogs/userlog-reports.js
ofaletse: Sure as per stant-up meeting yesterday I indicated that webservers and sys controller needs to be changed. I have added katstore process in #CB-2758, thanks for bringing that to my attention.

app/userlogs/userlog-reports.js
```diff
@@ -3,16 +3,20 @@
...
 
             var vm = this;
             vm.startTime = new Date();
             vm.endTime = vm.startTime;
             vm.startDatetimeReadable = moment(vm.startTime.getTime()).format('YYYY-MM-DD 00:00:00');
             vm.endDatetimeReadable = moment(vm.endTime.getTime()).format('YYYY-MM-DD 23:59:59');
             vm.tags = UserLogService.tags;
-            vm.logFiles = UserLogService.logFiles;
+            vm.logFiles = ['activity', 'alarm'];
+            vm.nodeNames = [];
```
lvdheever: This is unused as you use a local `var nodenames` below.

ofaletse:
@lvdheever thanks for taking your time and review. Much appreciated.

PR #373, opened 2018-04-24
User/ofaletse/cb 2834/change/gui to handle elasticsearch queried logs

ofaletse:
@lvdheever Please review.

This PR handle the results from [PR](https://github.com/ska-sa/katportal/pull/331) i.e results from elasticsearch query interface.

The sceen grap below is to show that we are still able to generating report.
pdfff


ofaletse:
Note: I will generate and push dist files later.

app/services/userlog-service.js
```diff
@@ -57,6 +57,21 @@
             return deferred.promise;
         };
 
+
+        api.queryLogFiles = function (logFiles, startTime, endTime) {
```
lvdheever: Name now will be better if it is e.g. `querySystemLogs` as it does not query the files any longer.
But KatGUI does have functionality to show the system log files and that is still needed. So not sure that you should change queryLogFiles to query elastic. Would you need to add another api call instead querySystemLogs and keep queryLogFiles (or does the system log page on KatGUI use a different interface?)
Answering my own question - yes I think it does - it probably uses the tail_log_fileserver or whatever it is called.
Please confirm that, but then just the name change please.

app/services/userlog-service.js
```diff
@@ -57,6 +57,21 @@
...
+            var defer = $q.defer();
+            $http(createRequest(
+                'get', urlBase() + '/logs/' + logFiles + '/' + startTime + '/' + endTime)).then(
```
lvdheever: logFiles is an incorrect name now. Please change to something else - programNames?

app/userlogs/userlog-reports.js
```diff
@@ -123,22 +123,22 @@
...
-                                    }
+                                    logLines.push({line: line});
                                 });
                                 pdf.autoTable(columns, logLines, {
                                     startY: pdf.autoTableEndPosY() + 50,
                                     theme: 'striped',
                                     margin: {top: 8, bottom: 8},
                                     overflow: 'linebreak'});
-                            });
+                                    logLines = [];
```
ajoubertza: This looks like it is indented a bit too far.  Should align with `pdf`

ofaletse: Not really. I will change code so that the closing brace is clear.

app/health/config-health-view/config-health-view.js
```diff
@@ -7,7 +7,9 @@
                                   ConfigService, StatusService, NotifyService, $stateParams) {
 
         var vm = this;
-        vm.mapTypes = ['Pack', 'Partition', 'Icicle', 'Sunburst'];
+        // OJ Temporarily disable map types, details in CB-2770
+        //vm.mapTypes = ['Pack', 'Partition', 'Icicle', 'Sunburst'];
+        vm.mapTypes = ['Sunburst'];
```
lvdheever: This should work, and I assume it was tested to ensure no unforeseen side-effects.

Also - please test if it should be disabled in receptor-health.js as well (I think so).

ofaletse: Well, receptor-health is working as expected. i.e no need to disable other maps.

lvdheever: Great! Thanks for checking.

PR #374, opened 2018-05-19
use interface changed message to detact if we have to re-subscribe an…

ofaletse:
@lvdheever or @ajoubertza please review.

Cbf config health subscription occur during init and during websocket re-connection, if there are some sensors that are added to the component after init, we will fail to subscribe to the new sensors, this imply we will not get updates from the newly added sensors. The workaround(temporary solution) thennn was to: 

1. Refresh browser so that init can be called. 
2. Check if there are new sensors(diff) and subscribe during sensorUpdate event, the problem with this implementation is that if there is no sensor update then subscription of new sensor will not be called. 

Clearly the above two workaround proved not to be ideal, to resolve the above case, the following solution(this PR) was adopted:
 
* Detects when an interface changed message is received(number of sensors changed) Note the interface changed we listening to is a NATS subject that we subscribed to.
* Unsubscribe to the pre-existing sensors.
* List all sensors that the component(cbf) have and subscribe again.

One of the change was to remove the hacking functions that were developed as a temporary measure(2). 
_As always the dist files will be added to after the PR is added._

JIRA: [CB-2819](https://skaafrica.atlassian.net/browse/CB-2819)


app/health/config-health-view/config-health-view.js
```diff
@@ -112,52 +109,9 @@
...
-                        sensors = vm.getDiffSensors(vm.subscribedSensors, vm.newList);
-                        sensors.forEach(function (sensor) {
-                            MonitorService.subscribeSensor(sensor);
-                            vm.subscribedSensors.push(vm.newList[sensor]);
-                        });
-                    }
-                }
-            });
-        };
+        var unbindReconnected = $rootScope.$on('interfaceChangedMessage', vm.initSensors);
```
lvdheever: I don't think you can re-use "unbindReconnected" here as it is used below for the websocketReconnected event.
Also - do you need to add "unbindInterfaceChanged" to Line 130 below with the other unbinds?

app/services/monitor-service.js
```diff
@@ -196,6 +196,10 @@
                         UserLogService.receivedUserlogMessage(msg.subject, data);
                     } else if (msg.subject.startsWith('portal.auth')) {
                         SessionService.receivedSessionMessage(msg.subject, data);
+                    } else if (msg.subject.startsWith('interface_changed')) {
+                        if (angular.isString(data) && data.startsWith('cbf')) {
+                            $rootScope.$emit('interfaceChangedMessage')
+                        }
```
lvdheever: I like this solution based on interface changed. However, not having followed the detail I have some questions.

It will be better if we do not restrict this to "cbf" only as it will be similar for any other Config Health view using any other component. If there is an interface changed on any component, then it will be good to handle that too in future. I am thinking that one could open a SYS config health display, then issue a "refresh-config" which may bring in a new antenna, and thus new SYS sensors and SYS will issue interface changed, or Subarray issues interface changed when it creates new precise time sensors. Although it is not the current use case, this generalisation will be good to build in now. 
Also then need to not make a single `interfaceChangedMessage` but keep the specification of which component issued the interfaceChangedMessage. That way the Config Health or other KatGUI forms can be more intelligent on whether they need to refresh its subscriptions depending on which component issued the interfaceChangedMessage. With the current implementation all the CBF displays for CBF1..4, CBF_DEV_1..4, CBF_MON_1..4 and CBFMON_DEV_1..4 will all refresh if any one of these CBF components issued an interface changed. We can then make the subscribers react to the `interfaceChangedMessage` more intelligently if the component is included in the emit.
For now, just make sure to include the component to the `interfaceChangedMessage` here and not restrict it to `startswith('cbf')`. We can then add the `startswith('cbf')` to config-health-view.js for now.

ajoubertza: There are many NATS messages starting with `interface_changed.`, which is the prefix we have subscribed to on line 151.  Besides normal KATCP interface changes, there is also a message for every sensor added and every sensor removed as part of the change. 

Some examples of the raw NATS messages are shown below (note: this is from a telnet session, before they are parsed, packaged and shipped over the websocket).
"plain" interface changed message:
```
MSG interface_changed.cbfmon_2 123 8
cbfmon_2
```
Add sensor message:
```
MSG interface_changed.add_sensor.cbf_roach_2.i0_state 123 316
{"status":"error","name":"cbf_roach_2_i0_state","component":"cbf_roach_2","original_name":"cbf_roach_2.i0-state","value":"disconnected","value_ts":1526888819.8865630627,"params":"['disconnected', 'syncing', 'synced']","time":1526888819.8874249458,"type":"discrete","description":"State of the device handler for i0"}
```
Remove sensor message:
```
MSG interface_changed.remove_sensor.cbfmon_2.i0_xhost01_network_cnt 123 80
{"escaped_name":"i0_xhost01_network_cnt","component":"cbfmon_2"}
```

Are you doing the `isstring` check to try and filter out the "plain" changes?

ofaletse: Yes. I am filtering out all the 'object' type messages(interface_changed.remove_sensor and interface_changed.add_sensor).

ofaletse: Please see the commit, as per our verbal discussion.

app/services/monitor-service.js
```diff
@@ -197,8 +197,8 @@
                     } else if (msg.subject.startsWith('portal.auth')) {
                         SessionService.receivedSessionMessage(msg.subject, data);
                     } else if (msg.subject.startsWith('interface_changed')) {
-                        if (angular.isString(data) && data.startsWith('cbf')) {
-                            $rootScope.$emit('interfaceChangedMessage')
+                        if (angular.isString(data)) {
```
lvdheever: May be good to add a comment here that you filter out interfaceChangedMessage.addSensor and interfaceChangedMessage.removeSensor objects so that only interfaceChangedMessage.<component> are emitted.

PR #378, opened 2018-05-31
User/bulelani/cb 2897/change maintenance blocks background

bxaia:
Changed the border to be smaller as opposed to what the name of the branch suggests. The background of the maintenance blocks remains showing the status of the particular sensor/device/antenna.

@ofaletse @lvdheever please review


app/health/health.less
```diff
@@ -69,14 +69,9 @@
...
-    stroke: #9C27B0 !important;
-    color: #fff;
-    border: 8px solid #9C27B0 !important;
-    text {
-        stroke: transparent;
-        fill: white;
-        font-weight: bold;
-    }
+    border: 2px solid #9C27B0;
+    background-color: #9C27B0;
```
lvdheever: Did we not say that the backbround-color will still be the status color and not purple? Does this change do that?

bxaia: The background-color will still remain the status color, this change here is only applied on the legend as legend was now showing a nominal state colour isntead of the purple colour. The code that actually keeps the background-color to be status color is in health.js line 73 - 89, pasted below:

```  
            if (sensorValue) {
                var inMaintenance = vm.sensorValues.katpool_resources_in_maintenance;
                if (inMaintenance && (
                    inMaintenance.value.indexOf(sub.sensor.replace('agg_', '').split('_')[0]) > -1 ||
                    inMaintenance.value.indexOf(sub.component) > -1)) {
                    statusClassResult.push('in-maintenance-child');
                }
                statusClassResult.push(sensorValue.status + '-item');
```

lvdheever: Tx

PR #379, opened 2018-06-05
User/ofaletse/cb 3016/add maintenance receptor health

ofaletse:
@lvdheever please review

This PR adds receptor health view requested by maintenance team. See the details below

[Maintenance Health.pdf](https://github.com/ska-sa/katgui/files/2070892/Maintenance.Health.pdf)


lvdheever:
@ofaletse Please add this PDF to the katcamconfig PR. The comments I made there (like DIG should not be under AP and adding some other bands) can be verified against this PDF. Please make sure that the content in the PDF is reflected in the katcamconfig PR.

PR #381, opened 2018-06-22
subscribe to messege portal.cache_update_done and process the data th…

ofaletse:
@ajoubertza please review

This PR subscribe to message _portal.cache_update_done_ and process the data that is received by determining if the message is from _cbfmon_ first and then subscribe to the sensors.

I tested the entire chain that proved to be working.

[Jira](https://skaafrica.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=CB&modal=detail&selectedIssue=CB-2819&quickFilter=12)

### DIFF
--- a/app/health/config-health-view/config-health-view.js
+++ b/app/health/config-health-view/config-health-view.js
@@ -109,8 +109,8 @@
             $rootScope.$emit('redrawChartMessage', {size: vm.treeChartSize});
         };
 
-        var handleInterfaceChanged = $rootScope.$on('interfaceChangedMessage', function (event, component) {
-            if (component.startsWith('cbfmon')) {
+        var handleInterfaceChanged = $rootScope.$on('portalCacheUpdateDone', function (event, data) {
+            if (JSON.parse(data.component).component.startsWith('cbfmon')) {
                 vm.initSensors();
             }
         });
--- a/app/health/config-health-view/config-health-view.js
+++ b/app/health/config-health-view/config-health-view.js
@@ -110,7 +110,7 @@
         };
 
         var handleInterfaceChanged = $rootScope.$on('portalCacheUpdateDone', function (event, data) {
-            if (JSON.parse(data.component).component.startsWith('cbfmon')) {
+            if (data.component.startsWith('cbfmon')) {
                 vm.initSensors();
             }
         });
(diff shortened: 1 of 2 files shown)


app/health/config-health-view/config-health-view.js
```diff
@@ -109,8 +109,8 @@
             $rootScope.$emit('redrawChartMessage', {size: vm.treeChartSize});
         };
 
-        var handleInterfaceChanged = $rootScope.$on('interfaceChangedMessage', function (event, component) {
-            if (component.startsWith('cbfmon')) {
+        var handleInterfaceChanged = $rootScope.$on('portalCacheUpdateDone', function (event, data) {
+            if (JSON.parse(data.component).component.startsWith('cbfmon')) {
```
ajoubertza: Fix the content of the published message in https://github.com/ska-sa/katportal/pull/332 (i.e. just sending component, instead of all the other data), will require a change here.  We won't need to specify `.component` twice.

app/health/config-health-view/config-health-view.js
```diff
@@ -109,8 +109,8 @@
             $rootScope.$emit('redrawChartMessage', {size: vm.treeChartSize});
         };
 
-        var handleInterfaceChanged = $rootScope.$on('interfaceChangedMessage', function (event, component) {
-            if (component.startsWith('cbfmon')) {
+        var handleInterfaceChanged = $rootScope.$on('portalCacheUpdateDone', function (event, data) {
+            if (data.component.startsWith('cbfmon')) {
```
lvdheever: I am not sure that we still need this restricted to cbfmon only - ideally we want to handle the portal cache update done event correctly for all components on all displays. Although the cgfmon may be the only use case we have encountered yet.
E.g. what if the ANC config health display is opened before the ANC got all its sensor registered (I'm thinking of some sensors like ganglia metrics which may be "late") or subarray manager that add some sensors on activation.
Now that we properly handle the fact that the sensor list cache update is done we should be able to do this for all components for which this display has sensors. But this work can be done in subsequent PR.

ofaletse: I hear you @lvdheever, and you are right. Reality is that calling _vm.initSensors()_ is very expensive. So yes lets handle it in the other PR's.

PR #383, opened 2018-06-28
config health changes for sunburst.

bxaia:
GUI changes for config health work to function properly.
Below is the screen shot of the test page. 

screen shot 2018-06-26 at 21 25 45


app/d3/receptor-health-sunburst-map.js
```diff
@@ -108,9 +108,14 @@ angular.module('katGui.d3')
                             var prefix = d.prefix? d.prefix : '';
                             var classStr = '';
                             var dataName = '';
+                            debugger;
                             if (scope.dataMapName instanceof Object) {
                                 dataName = d.component + '_' + d.sensor;
-                                classStr = dataName;
+                                if (d.component == "all") {
+                                  classStr = dataName = d.sensor;
+                                } else {
```
lvdheever: Not important, but it will be clearer to put line 113 with the "dataName" assignment in the 'else' statement here.

PR #384, opened 2018-06-29
now links pointing to logtrail

ofaletse:
@bxaia  or @lvdheever 
Please review, This PR creates links that point to logtrail
screen shot 2018-06-29 at 8 32 16 am

No Jira


app/app.js
```diff
@@ -422,28 +422,18 @@
                 kibanaUrl = [
                     "http://",
                     ConfigService.systemConfig.system.kibana_server,
-                    "/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:30000),",
-                    "time:(from:now-10m,mode:relative,to:now))&",
```
ajoubertza: Do we now configure the amount of history to display on logstash instead of as part of the URL?
How much history comes up by default?

ajoubertza:
It will be good to kat-update devlab after this so that the logs are working again.  I'll make a ticket for the next sprint.

app/app.js
ofaletse: We now set this in logtrail config, the default is 1 day. Might be too much but the latest logs are  displayed first. It looks ok according to me.

ofaletse:
Okay, will give it a try now after merging.

lvdheever:
Logtrail looks great but I am a bit worried with this merge to master because it means we have broken it for the RTS upgrade in the coming sprint. We do not want to run different Elastic clusters for RTS and MKAT and Ofaletse said that the logtrail upgrade was not great. If we need to do the logtrail upgrade for RTS it will affect MKAT. 

I have made a ticket in the RTS epic to resolve this. Ofaletse, please start thinking about what is needed to test in camlab to resolve this before the RTS Karoo upgrade and add that to the ticket.

ofaletse:
@lvdheever Good point Lize. I will start thinking of this.
@gtrymore when you get a second may you please, get the lab server ready, I would like to add ELK Cluster.

PR #385, opened 2018-07-13
receptorHealth expect a parameter, also include maintenance button in…

ofaletse:
@lvdheever please review.
This PR fix the navigation widget to take a parameter(maintenance or receptor) when calling _receptorHealth_, furthermore add the maintenance button to the navigation widget.


ofaletse:
@lvdheever, Will do.

PR #386, opened 2018-07-16
User/ofaletse/hotfix add maintenance to navigation widget and change the way it is called

ofaletse:
@lvdheever please review

This PR is a hotfix to v20,  from #385 
It fix the navigation widget to take a parameter(maintenance or receptor) when calling receptorHealth, furthermore add the maintenance button to the navigation widget.


ofaletse:
Thanks for reviewing.

PR #388, opened 2018-07-25
User/ofaletse/make aggregate sensor to be updated

ofaletse:
@lvdheever or @bxaia please review, this PR makes it possible to update the aggregate sensors.
See below the attached image.
screen shot 2018-07-25 at 3 08 37 pm


app/health/config-health-view/config-health-view.js
```diff
@@ -120,6 +120,13 @@
             if (!view || sensor.name.search(view.sensors.join('|')) < 0) {
```
ajoubertza: Not sure how all of this fits together, but the search (for filtering?) will still use the original name, e.g. with the mon_monctl prefix.  Is that what we want?

ofaletse: Yes that true. Because of ```!view``` we will continue.

ofaletse:
Thanks

PR #390, opened 2018-08-21
Create user log from schedule block and subarray

ofaletse:
@bxaia or @ajoubertza please review.

This PR allows the user to creates user log from schedule block and also from subarray.

[JIRA](https://skaafrica.atlassian.net/browse/MT-12)

### DIFF
--- a/app/scheduler/scheduler-home.js
+++ b/app/scheduler/scheduler-home.js
@@ -998,7 +998,9 @@
         vm.getResources = function (spec, dry_run_alloc, alloc) {
             resources = [];
             if (spec) {
-                resources = spec.split(',')
+                if (!spec.indexOf('{') > -1) {
+                    resources = spec.split(',')
+                }
             }
             if (dry_run_alloc) {
                 resources = dry_run_alloc.split(',');
@@ -1015,8 +1017,8 @@
             var allocations = [];
             var assignedResources = [];
             var start_time = $rootScope.utcDateTime;
-            dryrun_link = ConfigService.GetKATTaskFileServerURL() + "/tailtask/" + sb.id_code + "/dryrun"
             if (sb) {
+                dryrun_link = ConfigService.GetKATTaskFileServerURL() + "/tailtask/" + sb.id_code + "/dryrun"
                 content = "Schedule block: " + sb.id_code + "\n\nDry run link: " + dryrun_link;
                 if (sb.actual_start_time) {
                     start_time = sb.actual_start_time.split('.')[0];
@@ -1027,20 +1029,19 @@
                 allocations = vm.getResources(sb.antenna_spec, sb.antennas_dry_run_alloc,
                     sb.antennas_alloc).concat(vm.getResources(sb.controlled_resources_spec,
                       sb.controlled_resources_dry_run_alloc, sb.controlled_resources_alloc));
-
-                for (var id = 0; id < allocations.length; id++) {
-                    if (_.findWhere(UserLogService.tags, {name: allocations[id]})) {
-                        assignedResources.push(
-                            _.findWhere(UserLogService.tags, {name: allocations[id]}));
+                for (var i = 0; i < allocations.length; i++) {
+                    var tag = _.findWhere(UserLogService.tags, {name: allocations[i]})
+                    if (tag) {
+                        assignedResources.push(tag);
                     }
                 }
             }
             else {
                 allocations = ObsSchedService.sensorValues[vm.subarray.name + '_allocations'].parsedValue;
-                for (var id = 0; id < allocations.length; id++) {
-                    if (_.findWhere(UserLogService.tags, {name: allocations[id][0]})) {
-                        assignedResources.push(
-                            _.findWhere(UserLogService.tags, {name: allocations[id][0]}));
+                for (var i = 0; i < allocations.length; i++) {
+                    var tag = _.findWhere(UserLogService.tags, {name: allocations[i][0]})
+                    if (tag) {
+                        assignedResources.push(tag);
                     }
                 }
             }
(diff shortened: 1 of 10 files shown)


app/scheduler/scheduler-home.js
```diff
@@ -995,6 +995,66 @@
             }
         };
 
+        vm.getResources = function (spec, dry_run_alloc, alloc) {
+            resources = [];
+            if (spec) {
+                resources = spec.split(',')
```
ajoubertza: Antenna specs  are a little tricky.  Could be a simple CSV, but can also be dict as a JSON string, so this isn't sufficient for the JSON case.  Easiest option is probably to skip the spec if it doesn't look like a CSV.  E.g. `{`characters would be a bad sign.

katpool has the details.  https://github.com/ska-sa/katobs/blob/343c00f/katobs/katpool/katpool.py#L687
Example usage:  https://github.com/ska-sa/katscripts/blob/0f390c0/integration_tests/tests/observation_control/test_subarray_control.py#L674-L675

app/scheduler/scheduler-home.js
```diff
@@ -995,6 +995,66 @@
...
+                content = "Schedule block: " + sb.id_code + "\n\nDry run link: " + dryrun_link;
+                if (sb.actual_start_time) {
+                    start_time = sb.actual_start_time.split('.')[0];
+                }
+                if (sb.actual_end_time) {
+                    end_time = sb.actual_end_time.split('.')[0];
+                }
+                allocations = vm.getResources(sb.antenna_spec, sb.antennas_dry_run_alloc,
+                    sb.antennas_alloc).concat(vm.getResources(sb.controlled_resources_spec,
+                      sb.controlled_resources_dry_run_alloc, sb.controlled_resources_alloc));
```
ajoubertza: Nice idea to use a function and join the results together.

app/scheduler/scheduler-home.js
```diff
@@ -995,6 +995,66 @@
...
+            return resources;
+        }
+
+        vm.addUserLog = function ($event, sb=null) {
+            var content = '';
+            var end_time = '';
+            var allocations = [];
+            var assignedResources = [];
+            var start_time = $rootScope.utcDateTime;
+            dryrun_link = ConfigService.GetKATTaskFileServerURL() + "/tailtask/" + sb.id_code + "/dryrun"
```
ajoubertza: This should be inside the `if (sb)` check.

app/scheduler/scheduler-home.js
```diff
@@ -995,6 +995,66 @@
...
+                    end_time = sb.actual_end_time.split('.')[0];
+                }
+                allocations = vm.getResources(sb.antenna_spec, sb.antennas_dry_run_alloc,
+                    sb.antennas_alloc).concat(vm.getResources(sb.controlled_resources_spec,
+                      sb.controlled_resources_dry_run_alloc, sb.controlled_resources_alloc));
+
+                for (var id = 0; id < allocations.length; id++) {
+                    if (_.findWhere(UserLogService.tags, {name: allocations[id]})) {
+                        assignedResources.push(
+                            _.findWhere(UserLogService.tags, {name: allocations[id]}));
```
ajoubertza: Why do we need to call `findWhere` twice with the same arguments?
Is `assignedResources` the indexes of the tags?  In that case, suggest changing the name to something that indicates this.

app/scheduler/subarrays-draft-assignment/subarrays-draft-assignment.html
```diff
@@ -98,7 +98,7 @@
                             <span class="icon-button fa fa-ellipsis-v" md-ink-ripple style="padding: 6px" ng-click="$mdMenu.open($event); $event.stopPropagation()"></span>
                             <md-menu-content>
                                 <md-menu-item>
-                                    <md-button ng-disabled="!(item.type === 'OBSERVATION' && parent.vm.iAmAtLeastCA())" ng-click="vm.verifyDraft(item)">Verify Draft</md-button>
+                                    <md-button ng-disabled="!(item.type === 'OBSERVATION' && parent.vm.iAmAtLeastCA())" ng-click="vm.verifyDraft(item)">Verify Assigned</md-button>
```
ajoubertza: This change is because the "Draft" schedule blocks are no longer visible on the GUI.  This menu option is only available on SBs that have been "Assigned" to the subarray.

app/userlogs/userlogs.js
```diff
@@ -211,15 +212,7 @@
                 content: '',
                 attachments: []
             };
-            vm.editUserLog(newUserLog, event);
-        };
-
-        vm.editUserLog = function (userlog, event) {
```
ajoubertza: The userlog-reports has a similar function.  Should it also use the new one at rootScope?
https://github.com/ska-sa/katgui/blob/558d875/app/userlogs/userlog-reports.js#L76
https://github.com/ska-sa/katgui/blob/b784051/app/userlogs/userlog-reports.html#L106

app/scheduler/scheduler-home.js
ofaletse: Thanks for that heads up.

app/scheduler/scheduler-home.js
ofaletse: Your right

app/userlogs/userlogs.js
ofaletse: I opt not to use the ```$rootScope``` one. Besides it looks like they are not doing the same thing although they are named in the same way.

ajoubertza:
It would be nice to have this on site, at least for MKAT, so can we get a v20 hotfix please?

ofaletse:
Good point, Will do, Today is engineering day. as per our verbal discussion will add to v21

ajoubertza:
Still needed the dist files to be updated...

ofaletse:
Yes. I will create a separate branch.

PR #391, opened 2018-08-22
add dist files and avoid passing null to a variable.

ofaletse:
@ajoubertza or @bxaia please review.


ofaletse:
self merging, not that critical for review....

PR #395, opened 2018-09-06
add a feature to be able to add user log from sensor graph and sensor…

ofaletse:
Add a feature that will allow the user to be able to open and add user log from sensor list display and sensor graph display. A context menu is added to the table of sensor graph and sensor list.
To archive this an angular right click and context menu library was used.

@bxaia and @ajoubertza may you please review.

Jira [MT-23](https://skaafrica.atlassian.net/browse/MT-23) 
Jira [MT-11](https://skaafrica.atlassian.net/browse/MT-11)


ajoubertza:
@ofaletse Can you please add some screenshots?

bxaia:
@ofaletse when clicking on sensor does it give you an option to pre-populate the compound tag with the sensor name

ofaletse:
Add user log from a graph (Sensor graph)
screen shot 2018-09-07 at 8 54 03 am

Compound tags added automatically (Sensor graph)
screen shot 2018-09-07 at 8 55 00 am

Proof that it is added (Sensor graph)
screen shot 2018-09-07 at 8 55 43 am

Proof that it is added during editing (Sensor graph)
screen shot 2018-09-07 at 8 56 13 am

Compound tags added automatically (Sensor list)
screen shot 2018-09-07 at 10 34 43 am

app/sensor-list/sensor-list.js
```diff
@@ -73,6 +73,68 @@
             vm.showValueTimestamp = $localStorage.sensorListShowValueTimestamp;
         }
 
+        vm.selectedSensor = '';
+        vm.getSelectedSensor = function(sensor) {
+            vm.selectedSensor = sensor;
+        }
+
+        vm.deriveCompoundTag = function(sensor) {
```
ajoubertza: Would be nice if this function could be shared between sensor list and sensor graph - less code to maintain.

app/sensor-graph/sensor-graph.js
```diff
@@ -38,6 +39,67 @@
...
+
+        vm.openUserLog = function() {
+          UserLogService.listTags();
+          var content = '';
+          var endTime = '';
+          var allocations = [];
+          var assignedResources = [];
+          var compoundTags = []
+          var startTime = $rootScope.utcDateTime;
+          if (vm.selectedSensor) {
```
ajoubertza: Does it really matter if a sensor is selected on the sensor graph page.  Isn't a check on  the length of `vm.sensorNames` sufficient?

app/sensor-list/sensor-list.js
ofaletse: Good point, But its a bit hard-ish.
This is because 
   Sensor graph: ```sensor.attributes.katcp_name``` and 
   Sensor list: ```sensor.original_name.```

app/sensor-graph/sensor-graph.js
ofaletse: I agree

PR #398, opened 2018-09-11
Add a right click library

ofaletse:
Yarn need to know the  the versions of each dependency that are installed.


ofaletse:
Thanks for the comment/review. Yes it does.
I added the dist in the other [PR](https://github.com/ska-sa/katgui/pull/399)

PR #400, opened 2018-10-04
User/tockards/mt 153/add value time to csv

tockards:
- add value_time to all queries to allow CSV download to include it.
- include min_value_time and value_time in CSV output.

### DIFF
--- a/app/d3/multi-line-chart.js
+++ b/app/d3/multi-line-chart.js
@@ -874,18 +874,21 @@ angular.module('katGui.d3')
                         var dataString = '';
                         var sensorInfo = sensorValues.values[i];
                         if (useUnixTimestamps) {
-                            if (includeValueTimestamp && (sensorInfo.value_time || sensorInfo.value_ts)) {
-                                dataString += ((sensorInfo.value_time || (sensorInfo.value_ts / 1000))) + ',';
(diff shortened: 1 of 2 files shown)


tockards:
@SKAJohanVenter previously the name of `value_time`  was `value_ts` and we changed that to 'value_times' and then we added `min_value_time` so its inconsistencies in the Katstore API. This is more of a fix and than a solution.

app/d3/multi-line-chart.js
```diff
@@ -874,18 +874,21 @@ angular.module('katGui.d3')
...
+                            dataString += ((sensorInfo.sample_time || sensorInfo.sample_ts / 1000 || sensorInfo.min_sample_time)) + ',';
                         } else {
                             var value_time = null;
                             var sample_time = null;
                             if (sensorInfo.value_time) {
                                 value_time = sensorInfo.value_time * 1000;
                             }
+                            if (sensorInfo.min_value_time) {
+                                value_time = sensorInfo.min_value_time * 1000;
+			    }
```
ofaletse: This brace looks a bit odd, for my eyes.

tockards: @ofaletse  fixed up the tab and space indentation.

PR #401, opened 2018-10-18
Change the usability report to use the new katstore

ofaletse:
@tockards please review. 
This PR add a functionality that make the usability report to be able to use the new katstore.
[Jira MT-138 ](https://skaafrica.atlassian.net/browse/MT-138)

### DIFF
--- a/app/reports/utilisation-report.js
+++ b/app/reports/utilisation-report.js
@@ -309,21 +309,20 @@
                 vm.creatingSubarrayReport = true;
 
                 DataService.sampleValueDuration(vm.subarrayReportSensorsRegex, startDate, endDate).then(function (result) {
-                    if (result.data) {
-
-                        result.data.forEach(function (item) {
+                    if (result.data.data) {
+                        result.data.data.forEach(function (item) {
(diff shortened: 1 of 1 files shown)


app/reports/utilisation-report.js
```diff
@@ -429,8 +427,7 @@
                                         vm.poolResourcesAssignedDurations[resource].durationTotal = vm.durationToString(duration);
                                         vm.poolResourcesAssignedDurations[resource].percentageTotal = vm.percentageOfTotalToString(vm.poolResourcesAssignedDurations[resource].durationTotalSeconds);
                                     }
-
-                                    if (!vm.poolResourcesAssignedDurations[resource][key].value) {
+                                    if (!vm.poolResourcesAssignedDurations[resource][key]) {
```
tockards: @ofaletse why the removal of `value` from `vm.poolResourcesAssignedDurations[resource][key]` ?
If its not there is False by default?

ofaletse: Some objects do not have ```value```

PR #402, opened 2018-10-19
User/ofaletse/mt 184/add sb compound tag

ofaletse:
@bxaia or @ajoubertza please review

This PR  automatically adds compound tag when the user creates user log from schedule block display.
This was a request from Rupert.
See screen shot
 
[Jira ](https://skaafrica.atlassian.net/browse/MT-184)


app/scheduler/scheduler-home.js
```diff
@@ -1020,6 +1020,7 @@
             if (sb) {
                 dryrun_link = ConfigService.GetKATTaskFileServerURL() + "/tailtask/" + sb.id_code + "/dryrun"
                 content = "Schedule block: " + sb.id_code + "\n\nDry run link: " + dryrun_link;
+                compoundTag = ["SB_:" + sb.id_code + ":_"]
```
ajoubertza: The separator for compound tag fields is `_:_`, so we seem to be missing some underscores here?

ofaletse:
@ajoubertza You are right, but this change was as per Rupert's email:
```
Thanks for the features :-)

Please tag the log on the schedule block with: _:{ ... }:_
Schedule_Block_:{1234556-1234}:_
or
SB_:{1234556-1234}:_
(or your preferred key word.) 
```

ajoubertza:
@ofaletse That still seems strange - maybe it was a typo.  Please can you confirm with him?

PR #404, opened 2018-10-22
New GUI page for SKARAB health

xinyuwu:
New page to support planned SKARAB health sensors which are not yet available. Dummy sensor is used. This feature needs to be modified and tested once SKARAB health sensors are available.

image

image

JIRA: [MT-183](https://skaafrica.atlassian.net/browse/MT-183)

### DIFF
--- a/app/d3/skarab-health-view.js
+++ b/app/d3/skarab-health-view.js
@@ -1,0 +1,198 @@
+angular.module('katGui.d3')
+
+.directive('skarabHealthView', function($rootScope, $timeout, d3Util) {
+    return {
+        restrict: 'E',
+        controller: 'SKARABHealthCtrl',
+        controllerAs: 'vm',
+        scope: {
+        },
+        link: function(scope, element) {
+            const format = d3.time.format.utc("%Y-%m-%d %X");
+
+            var circles = null;
+
+            var margin = {
+                top: 20,
+                right: 20,
+                left: 20,
+                bottom: 50
+            };
+
+            var chart_margin = {
+              left: 30
+            }
+            var svg;
+            var width, height, radius;
+
+            scope.vm.redrawStatus = function () {
+
+              width = element[0].clientWidth-margin.left-margin.right;
+              height = width*scope.vm.NUM_OF_RACKS/scope.vm.NUM_OF_SLOTS;
+
+              radius = width/(scope.vm.NUM_OF_SLOTS*2)-5;
+              d3.select('svg').remove();
+
+              d3.select(element[0]).append("svg")
+                      .attr("width", element[0].clientWidth)
+                      .attr("height", height + margin.bottom)
+                      .attr("transform", "translate(0" + ", " + margin.top + ")");
+
+              svg = d3.select('svg');
+
+               var x = d3.scale.linear().domain([0,scope.vm.NUM_OF_SLOTS+1]).range([0, width]);
+               var y = d3.scale.linear().domain([0,scope.vm.NUM_OF_RACKS+1]).range([0, height]);
+
+               var xAxis = d3.svg.axis().scale(x).orient("top").ticks(10);
+               var yAxis = d3.svg.axis().scale(y).orient("right").ticks(10);
+
+              function draw_circle(selection) {
+                 var circleAttributes
+                     = selection.attr("cx", function (d) { return x(d.slot); })
+                                .attr("cy", function (d) { return y(d.rack); })
+                                .attr("ObjectID",function(d) {return d.id;})
+                                .style("stroke", "white")
+                                .style("stroke-width", 0)
+                                .on('mouseover',function(d) {
+                                  d3.select(this)
+                                 	  .style("fill-opacity", .8)
+                                    .style("stroke-width", 5);
+
+                                  let x =  d3.event.pageX;
+                                  if (d.slot > 43*2/3)
+                                    x = x-350;
+                                  tooltipdiv.html(
+                                        "<table>" +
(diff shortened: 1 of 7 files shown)


app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+            vm.data.push(obj);
+            id++;
+          }
+        }
+
+        ConfigService.getSystemConfig();
+
+        vm.initPositionSensors = function() {
+
+//          let sensorsRegex = 'skarab.position';
```
ajoubertza: Future work is required here when the real sensors are available.  Suggest adding a `TODO` comment to that describes our intention.

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+        vm.data = [];
+        var id = 0;
+        for (var i=1; i<(vm.NUM_OF_RACKS+1); i++) {
+          for (var j=1; j<(vm.NUM_OF_SLOTS+1); j++) {
+            var obj = {};
+            obj.rack = i;
+            obj.slot = j;
+            obj.position = '';
+            obj.sensor_name = '';
+            obj.sensor_value = {};
```
ajoubertza: Why are these variable names using underscores instead of camel case?

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+
+          var p = position.split(/[-,+]/);
+          var rack = Number(p[0].substring(1));
+          var slot = Number(p[1]);
+
+          for (var i=0; i<vm.data.length; i++) {
+            var obj = vm.data[i];
+            if (obj.rack==rack && obj.slot==slot) {
+              obj.position = position;
+              obj.sensor_name = sensorName;
```
ajoubertza: There could be two sensors linked to the same position (one each from `cbfhealth` and `cbfhealth2` proxies), but this mechanism won't allow for that.  The CBF monitoring spec record mentions that tie breaking will need to be done in this case to determine which of the two sensors to use.
We don't have to implement this right now, but at least add a TODO to remind us about it.

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+
+    function SKARABHealthCtrl($rootScope, $scope, ConfigService, StatusService, NotifyService,
+        MonitorService, d3Util, $timeout, $log) {
+
+        var vm = this;
+        vm.NUM_OF_RACKS = 9;
+        vm.NUM_OF_SLOTS = 42;
+
+        vm.subscribedSensors = [];
+        vm.sensorsRegex = '';
```
ajoubertza: This variable doesn't seem to be used?

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+        vm.initPositionSensors = function() {
+
+//          let sensorsRegex = 'skarab.position';
+          var sensorsRegex = 'device_status';
+          MonitorService.listSensorsHttp('all', sensorsRegex).then(function (result) {
+              for (var i=0; i<result.data.length; i++) {
+                var sensor = result.data[i];
+                var position = sensor.value;
+                position = 'B09-30+1';
+                var device_status_name = sensor.name.replace('position', 'device_status');
```
ajoubertza: Camel case naming?

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+              return;
+            }
+          }
+        };
+
+        vm.updateStatusSensor = function(sensor) {
+          for (var i=0; i<vm.data.length; i++) {
+            var obj = vm.data[i];
+            if (obj.sensor_name==sensor.name) {
+              obj.sensor_value = sensor;
```
ajoubertza: A sensor object has a value field, so this name is a bit confusing for me.  Suggest `obj.sensor = sensor` or `obj.sensorValue = sensor.value`.

app/health/skarab-health/skarab-health.js
```diff
@@ -0,0 +1,118 @@
...
+        var id = 0;
+        for (var i=1; i<(vm.NUM_OF_RACKS+1); i++) {
+          for (var j=1; j<(vm.NUM_OF_SLOTS+1); j++) {
+            var obj = {};
+            obj.rack = i;
+            obj.slot = j;
+            obj.position = '';
+            obj.sensor_name = '';
+            obj.sensor_value = {};
+            obj.status = 'unknown';
```
ajoubertza: I see github syntax highlights `status` - is it something special in JavaScript?

app/d3/skarab-health-view.js
```diff
@@ -0,0 +1,199 @@
...
+
+                       return radius;
+                    })
+                    .attr("class", function(d) {
+                      return d.status + "-child";
+                    });
+            };
+
+            scope.$on('$destroy', function() {
+                $interval.cancel(scope.unbindResize());
```
ofaletse: I could not locate ```$interval``` this might fail when we close the page.

app/health/skarab-health/skarab-health.js
ajoubertza: From our coding guidelines, we are supposed to be following this guide: https://github.com/toddmotto/angularjs-styleguide

app/health/skarab-health/skarab-health.js
xinyuwu: status is not a name you can't use. It is a name that should be avoided. In this case it is an attribute of obj and consistent with sensor.status, so it is safe.

ajoubertza:
Just remember to rebuild and commit the minified files.  I suggest waiting until @ofaletse has approved the PR.  Then we'll only need one more rebuild.

PR #407, opened 2018-11-09
Aligning Search TextBoxes on SensorGraph, 

bxaia:
Also addressing build errors that were problematic in the previous branch. I will delete the branch user/bulelani/MT-264/align_text_boxes as it does not build properly.

screen shot 2018-11-09 at 13 01 47
screen shot 2018-11-09 at 13 02 20


package-lock.json
```diff
@@ -0,0 +1,6276 @@
+{
```
ajoubertza: Can we remove this file or is it required?  If not required, please add it to the `.gitignore` file, so we don't end up committing it in future.

PR #412, opened 2018-11-21
Fix the percentages of resources used in sub-arrays

ofaletse:
@bxaia or @xinyuwu or @ajoubertza please review.

This PR fix utilization so that the percentages of resource used in subarrays sum up to the total sum of the resources.

Before
before

After
after

[Jira](https://skaafrica.atlassian.net/browse/MT-271)


ajoubertza:
Oh, in future please use the recently created feature branch: `feature/new-katstore-integration`, instead of this user branch.  That fits our branching model better.

PR #414, opened 2018-11-23
Fixing compound tags on sensor-list

bxaia:
Handling a case where sensors are divided with underscores, for formulation of compound tags on sensor-list page, where compound tags were not populating for sensors divided with underscores only.

Dist. files will be added once PR has been approved. Please find below screen-shots of the derived 

Jira: MT-295

screenshot 2018-11-23 at 15 58 37

screenshot 2018-11-23 at 15 59 08

### DIFF
--- a/app/sensor-list/sensor-list.js
+++ b/app/sensor-list/sensor-list.js
@@ -94,10 +94,23 @@
               " Time: " + sensor.received_timestamp +
               "\nValue: " + sensor.value
               startTime = sensor.timestamp
-              compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
-              if (compoundTag) {
+              if (sensor.original_name){
+                compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
+                if (compoundTag) {
                   compoundTags.push(compoundTag)
-              }
+                }
+              } else {
+                  _device_name = sensor.shortName.split(/_(.+)/)[0]
+                  _sensor = sensor.shortName.split(/_(.+)/)[1]
+                  if (_device_name && _sensor) {
+                    sensor.shortName = _device_name.concat('.', _sensor)
+                  }
+                  original_name = sensor.component.concat('.', sensor.shortName)
+                  compoundTag = $rootScope.deriveCompoundTag(original_name)
+                  if (compoundTag) {
+                    compoundTags.push(compoundTag)
+                  }
+                }
           }
           var tag = _.findWhere(
               UserLogService.tags,
--- a/app/sensor-list/sensor-list.js
+++ b/app/sensor-list/sensor-list.js
@@ -94,16 +94,16 @@
               " Time: " + sensor.received_timestamp +
               "\nValue: " + sensor.value
               startTime = sensor.timestamp
-              if (sensor.original_name){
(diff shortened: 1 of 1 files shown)


app/sensor-list/sensor-list.js
```diff
@@ -94,10 +94,23 @@
...
-              compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
-              if (compoundTag) {
+              if (sensor.original_name){
+                compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
+                if (compoundTag) {
                   compoundTags.push(compoundTag)
-              }
+                }
+              } else {
+                  _device_name = sensor.shortName.split(/_(.+)/)[0]
```
ajoubertza: What does the `_` prefix in this name indicate?  Also, we favour camel case naming in JavaScript.

ajoubertza: The first part of a sensor name on a component is not always a device, but it is a reasonable approximation, if we don't have any other info.

app/sensor-list/sensor-list.js
```diff
@@ -94,10 +94,23 @@
...
+                  _sensor = sensor.shortName.split(/_(.+)/)[1]
+                  if (_device_name && _sensor) {
+                    sensor.shortName = _device_name.concat('.', _sensor)
```
ajoubertza: Does this also modify the `sensor` object in the original container `vm.sensorValues` (see line 91)?  If so, should we be modifying it?

app/sensor-list/sensor-list.js
```diff
@@ -94,10 +94,23 @@
               " Time: " + sensor.received_timestamp +
               "\nValue: " + sensor.value
               startTime = sensor.timestamp
-              compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
-              if (compoundTag) {
+              if (sensor.original_name){
```
ajoubertza: Please add a space before the `{`

app/sensor-list/sensor-list.js
```diff
@@ -94,10 +94,23 @@
...
+                compoundTag = $rootScope.deriveCompoundTag(sensor.original_name)
+                if (compoundTag) {
                   compoundTags.push(compoundTag)
-              }
+                }
+              } else {
+                  _device_name = sensor.shortName.split(/_(.+)/)[0]
+                  _sensor = sensor.shortName.split(/_(.+)/)[1]
```
xinyuwu: Do you need to check if length of sensor.shortName.split(/_(.+)/) is two before doing _sensor = sensor.shortName.split(/_(.+)/)[0] and [1] to be sure?

app/sensor-list/sensor-list.js
bxaia: ok, thanks

app/sensor-list/sensor-list.js
ajoubertza: I disagree - the `sensor` object is a reference to the object in the map, so "both" get changed. 
For example:
image

In this case, no change is required since the `shortName` field isn't used anywhere else in the controller.

ajoubertza:
@bxaia Did you update the dist files first, before merging?

PR #423, opened 2019-01-17
Add navigation links in process control and sensor list

ofaletse:
@xinyuwu or @bxaia please review.

This PR makes navigation a bit easy
* Right-click a sensor from sensor-list and open sensor-graph.
* Add a functionality to be able to open a logtrail and navigate to sensor list from process control display.
* From process control display navigate easily to the sensor-list from the selected component/proxy.


ofaletse:
@xinyuwu Thanks for the review.

@ajoubertza Please come to my desk when ready for the exclusive wonderful demo. Just on a different  Xinyu and I are of the view that the dist files are pointless, to be committed especially because Xinyu will be adding two PR's soon. We of the view that Jenkins(tick) is sufficient to tell if one introduce a bug or not. I think it is counterproductive to be changing binary files that are not used. In case you have a different view please share, and I can add the dist files later if you perhaps see being valuable.

I will therefore merge this ticket(while your busy).

PR #424, opened 2019-01-18
User/xinyuwu/mt 344/update skarab health page

xinyuwu:
New skarab health page connected to skarab sensors (simulated)


ofaletse:
@xinyuwu Slots labels, are they not suppose to be the other way around? i.e ```05``` being at the top and ```40``` being at the bottom. The same applies to Rack labels.
Just asking, No need to change if it looks ok, in the eyes of the user and developer

app/d3/skarab-health-view.js
```diff
@@ -10,146 +10,172 @@ angular.module('katGui.d3')
...
+              height = element[0].clientHeight - margin.top - margin.bottom;
+
+              unitWidth = width/scope.vm.NUM_OF_RACKS - 2;
+              unitHeight = height/scope.vm.NUM_OF_SLOTS - 2;
 
-              radius = width/(scope.vm.NUM_OF_SLOTS*2)-5;
               d3.select('svg').remove();
 
-              d3.select(element[0]).append("svg")
+              var tooltipdiv = d3.select(element[0]).append("div")
```
ofaletse: It looks like you are no longer using the tool tip. Was this intended? Last time I check you where showing tool tip on a sensors that ware not nominal.

app/d3/skarab-health-view.js
```diff
@@ -10,146 +10,172 @@ angular.module('katGui.d3')
...
+                            })
+                            .on('mouseout',function (d) {
+                              d3.select(this)
+                               .style("fill-opacity", 1)
+                               .style("stroke-width", 0)
+                               .style("cursor", "default");
+                            })
+                            .on('click', function(d){
+                              if (d.name) {
+                                var components = d.name.split('\.');
```
ofaletse: This might fail(I think). You might consider handling a general exception here.

xinyuwu:
@ajoubertza <https://github.com/ajoubertza> double checked the skarab
health specification. Layout is the other way round from initial
implementation, ie B9 - B1 left to right, slot 42 to slot 1 top to bottom.
Replicating the exact hardware layout is top priority for skarab health
view, thus changed to match accordingly.

app/d3/skarab-health-view.js
```diff
@@ -10,146 +10,172 @@ angular.module('katGui.d3')
...
             var width, height, radius;
 
             scope.vm.redrawStatus = function () {
 
-              width = element[0].clientWidth-margin.left-margin.right;
-              height = width*scope.vm.NUM_OF_RACKS/scope.vm.NUM_OF_SLOTS;
+              width = element[0].clientWidth - margin.left - margin.right;
+              height = element[0].clientHeight - margin.top - margin.bottom;
+
+              unitWidth = width/scope.vm.NUM_OF_RACKS - 2;
```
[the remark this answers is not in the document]
xinyuwu: That's for the border between the rectangles.

app/d3/skarab-health-view.js
xinyuwu: Now that the text is displayed for all skarabs regardless of state and if you click on one it takes the user to the corresponding sensors, there does not seem to be need for tooltip. It can be restored easily if required.

app/d3/skarab-health-view.js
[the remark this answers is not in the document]
xinyuwu: defined a variable 'border' to make code more readable

app/d3/skarab-health-view.js
xinyuwu: Deleted tooltip div to make code less confusing.

PR #425, opened 2019-01-18
enable user to select min elevation on main plot

xinyuwu:
[MT-327:](https://skaafrica.atlassian.net/browse/MT-327) enable user to select min elevation on main plot
Screenshots in ticket.


bxaia:
@ajoubertza @ofaletse after chatting with the operators, there are few changes they requested on the skyplot display, these commits address that. They change the selectedMinimumElevation from 20 to 15 and also stepping with 5 on the Elevation Slider. 

The second commit is to fix a bug where antennas on the same subarray were not correctly allocated the same colour.

PR #427, opened 2019-01-28
User/bngcebetsha/MT-356/increase_throttlle_time_from300ms

BuntuN:
Throttle set to 10 000ms

This is to make katgui send fewer update requests when SBs are starting.
Main change is in `app/services/obs-sched-service.js`. The other files are a result of setting up katgui on my localhost, they should not affect anything and if they do please advice.
Requests are now 10s apart instead of 300ms.

JIRA: [MT-356](https://skaafrica.atlassian.net/browse/MT-356)


bxaia:
you can add the files @kmadisa  mentioned onto gitignore

BuntuN:
> The following files do not need to be committed:
> 
> * package.json
> * package-lock.json

PR #445, opened 2019-03-26
Hotfix v22 - Dryrun outside subarray 

BuntuN:
A v22 hotfix to include dryrun without a subarray in MeerKAT changes in `katgui`.

[MT-600](https://skaafrica.atlassian.net/browse/MT-600)


ajoubertza:
Looks about right.  Let's do some testing before merging.

PR #446, opened 2019-04-03
S-Band subband selection support in KATGUI

BuntuN:
Adding the option to select the centre frequency of the band on Gui

The Gui will have an option `f` where the centre frequency for each receiver band will be displayed and changed (in the case of a band with many centre frequencies). The bands with one centre frequency set it as default.

[MT-574](https://skaafrica.atlassian.net/browse/MT-574)

before
before2
before3
before4
before5
before6
before7
before8
before9
before10
before11
before12
before13


app/scheduler/scheduler-home.js
```diff
@@ -123,6 +126,24 @@
...
             });
 
+
+        ConfigService.getSubBandConfig()
+            .then(function(subBandConfig) {
+                vm.sub_bands = [];
+                vm.subBandsMap = {};
+                var subBandKeys = Object.keys(subBandConfig);
+                subBandKeys.forEach(function(sub_band) {
+                      vm.sub_bands.push({
```
ajoubertza: This variable is not used anywhere, so it should be removed.

BuntuN:
Fixed dropdown menu to show MHz 

Screen Shot 2019-04-03 at 1 43 43 PM

PR #448, opened 2019-04-12
Show index position on subarray activation page

bxaia:
This pull request is adding a display element that uses the `ridx-position` sensor from each receptor.  The sensor has possible values like “x”, “u”, “s”, “l”, “moving” and “undefined”.   Attached are screenshots of two subarrays from site which have different Index positions.

Screenshot 2019-04-12 at 15 16 34
Screenshot 2019-04-12 at 15 05 23


app/scheduler/scheduler.less
```diff
@@ -753,3 +753,19 @@
     overflow-x: auto;
     overflow-y: hidden;
 }
+
+.subarray_resources .nominal {
+    color: black;
+}
+
+.subarray_resources .moving {
```
ajoubertza: `moving` is not an option for a sensor status.   More likely `warn`.

app/scheduler/scheduler.less
```diff
@@ -753,3 +753,19 @@
...
+    color: orange;
+}
+
+.subarray_resources .error {
+    color: red;
+}
+
+.subarray_resources .undefined {
```
ajoubertza: This is also not a KATCP sensor status. Here are the options, from the KATCP guidelines:  https://katcp-python.readthedocs.io/en/latest/_downloads/NRF-KAT7-6.0-IFCE-002-Rev5.pdf

image

PR #452, opened 2019-04-23
speed up observation schedule db requests to improve gui performance

xinyuwu:
[MT-655](https://skaafrica.atlassian.net/browse/MT-655)


ajoubertza:
We just need to wait for ska-sa/katportal/pull/368 before merging this one.

PR #458, opened 2019-05-29
Aligning values in the weather display

bxaia:
This pull request is aligning the sensor name and values on the weather display and also overriding the minimum width inherited from the resource-sensor-value-item class, which caused the shifting of the "MAX" and "MAX at" values out of the display.

JIRA: MT-705

Screenshot 2019-05-29 at 15 29 14


mmphego:
:+1: Nice.

PR #461, opened 2019-06-03
Hotfix v22: GUI improvements on Weather and Edit User Displays and updating Jenkinsfile

bxaia:
This PR are fixes to the Weather display, properly aligning the MAX and MAX at sensor values. This also includes the correct wording on the Edit User Dialog Box, from "Edit User" to "SAVE" after editing the user info.

Jira: MT-705, MT-706


dist/version.txt
```diff
@@ -1 +1,10 @@
+<<<<<<< HEAD
+<<<<<<< HEAD
```
ajoubertza: There are git conflicts here, and duplicate minified files - that should be fixed after the dist files are rebuilt.

bxaia:
I have just added the dist file which has fixed the git conflicts and duplicate files.

PR #462, opened 2019-06-26
GUI prevents non compatible resources to subarray

bxaia:
This PR restricts the resources that are not compatible to specific subarray to be assigned to it by hiding them from being displayed on the subarray.  

Dist files will be compiled when PR is approved.

Jira: [MT-654](https://skaafrica.atlassian.net/browse/MT-654)

Screenshot 2019-06-26 at 15 54 59
Screenshot 2019-06-26 at 15 54 45
Screenshot 2019-06-26 at 15 54 32
Screenshot 2019-06-26 at 15 54 16


app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
             ObsSchedService.assignResourcesToSubarray(vm.subarray.id, resourceName);
         };
 
+        vm.resourceAllowedInSubarray = function (resourceName) {
+            var specificResources = [];
+            var generic_resources = ConfigService.systemConfig['internals']['multiples'].split(',');
+            var defaultResources = ['anc', 'ptuse_1', 'ptuse_2'];
+            for (var i=0; i < generic_resources.length; i++) {
+                var specificResource = generic_resources[i] + (vm.subarray.id);
```
ofaletse: `ng-show` is called continuously... 
Sometimes the `vm.subarray` is undefined, this will imply that `vm.subarray.id` will through an error, please handle that case in your code.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
             ObsSchedService.assignResourcesToSubarray(vm.subarray.id, resourceName);
         };
 
+        vm.resourceAllowedInSubarray = function (resourceName) {
+            var specificResources = [];
+            var generic_resources = ConfigService.systemConfig['internals']['multiples'].split(',');
+            var defaultResources = ['anc', 'ptuse_1', 'ptuse_2'];
```
ajoubertza: We should be able to implement without hardcoding any of these.  Same for receptors.  If the name doesn't start with any of the options in generic resources, then it is allowed in the subarray.

PTUSE will increase to 4 instances at some point in future, so at least extend this to include ptuse_3 and ptuse_4.  If you must stick with hardcoding.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
...
+            for (var i=0; i < generic_resources.length; i++) {
+                var specificResource = generic_resources[i] + (vm.subarray.id);
+                specificResources.push(specificResource);
+            }
+            if (specificResources.includes(resourceName) || resourceName.startsWith('m0') ||
```
ajoubertza: This will hide for SKA MPI dishes.  Resources could also start with `'s0'`

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
...
+                defaultResources.includes(resourceName) || resourceName.endsWith(vm.subarray.id)) {
```
ajoubertza: Why do we need this final check `resourceName.endsWith(vm.subarray.id)`?  What case would use it?

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,22 @@
...
+            if (vm.subarray) {
+                for (var i=0; i < generic_resources.length; i++) {
+                    var specificResource = generic_resources[i] + (vm.subarray.id);
+                    specificResources.push(specificResource);
+                }
+                if (specificResources.includes(resourceName) || resourceName.startsWith('m0') ||
+                    defaultResources.includes(resourceName) || resourceName.endsWith(vm.subarray.id)) {
+                    return true;
+                }
+            }
```
ajoubertza: Shouldn't this function return `false` if the resource is not allowed in the subarray?
Or are we relying on JavaScript to evaluate no return value as false?

bngcebetsha: returning `false` hides everything

app/scheduler/subarray-resources/subarray-resources.js
ajoubertza: @ofaletse Can you update your review, please?  This issue has been addressed.

PR #466, opened 2019-07-03
Hotfix V22-Prevent assigning non-compatible resources to subarray

bngcebetsha:
Prevent assigning generic resources `[bluse_, cbf_, cbf_dev_, fbfuse_, sdp_, tuse_]` to subarrays where they don't belong, e.g building a subaaray with `sdp_1` on `subarray_3`. On ech subarray will be visible resources relevant to that subarray. The `receptors`, `ptuse_1`, `ptuse_2, ...`, and  `anc` resources will be available in all subarrays.

[MT-654](https://skaafrica.atlassian.net/browse/MT-654)


app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,27 @@
             ObsSchedService.assignResourcesToSubarray(vm.subarray.id, resourceName);
         };
 
+
+        vm.resourceAllowedInSubarray = function (resourceName) {
+            var genericResources = [];
+            var generic_to_specific_resources = ConfigService.systemConfig['internals']['generic_to_specific_resources'].split(',');
```
tockards: LGTM!

For interest sake @xinyuwu @ajoubertza 
What is the preference of camelCase and using underscores with regards to variables. 
A quick glance of https://github.com/toddmotto/angularjs-styleguide  does not mention much on syntactic conventions for variables.

ajoubertza: Good point, @tockards.  We haven't been so strict in the GUI, but should be more so.

I like consistency.  In JavaScript, the default should be camelCase, since that is what is used most often.  The only exception is when communicating with portal - e.g. loading data from JSON blobs - those will often have underscore names due to the Python code/sensor names.

ajoubertza: In this case, `genericToSpecificResources` would have been a better name, but no need to change now.

PR #489, opened 2019-09-17
Sensor graph issue with undefined value

xinyuwu:
[MT-1000](https://skaafrica.atlassian.net/browse/MT-1000)

Screenshot 2019-09-17 at 14 04 07


tockards:
1.) I am not a `GUI Developer` but is this fix and the ones that I did (see https://github.com/ska-sa/katgui/pull/486 )  bad design? (we are doing these changes on the `View` layer to my understanding)
2.) How do we not interpret a `0` result  as a `false positive`?

PR #490, opened 2019-09-19
Button to assign all usable receptors at once

bngcebetsha:
1. This feature allows users to add all receptors to the subarray at the click of the purple button. All receptors in maintenance and marked faulty will not be assigned.

2. A subarray in maintenance will allow all receptors, including those marked faulty and in maintenance to be assigned.

3. The blue button has changed tooltip from `Assign Resources to Subarray` to `Assign Selected` and the new button(purple) says `Assign All Receptors`

4. The new button adds only receptors, the user must select the data proxies manually.

[MT-979](https://skaafrica.atlassian.net/browse/MT-979)

Untitled

Untitled2

untitled3

Untitled4


app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
             ObsSchedService.assignResourcesToSubarray(vm.subarray.id, resourceName);
         };
 
+        vm.assignAllReceptors = function (ants) {
```
ofaletse: I can't see where `ants` is used. and again where you are calling the function i think you are not passing it parameters

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -49,6 +49,20 @@
...
+            var configReceptors = ConfigService.systemConfig["antenna_labels"]["ALL"].split(',');
+            for (var i=0; i<configReceptors.length; i++) {
```
ofaletse: If you assign all available receptors(purple button) to subarray1 and then you go to another subarray3 and assign all available receptors(purple button) you are going to ask katpool to assign the resources, if this is site you will get 64 alerts message informing you that receptors are not available. Maybe you need to consider having a condition to check if there are free resources first. using this `vm.poolResourcesFree` might help you.

bngcebetsha: very good point Coach

PR #510, opened 2019-10-21
Add support for narrow band setting in GUI

xinyuwu:
[MT-1053](https://skaafrica.atlassian.net/browse/MT-1053)

Screenshot 2019-10-21 at 20 06 36

### DIFF
--- a/app/scheduler/templates/subarray-config-container.html
+++ b/app/scheduler/templates/subarray-config-container.html
@@ -72,7 +72,24 @@
             </md-button>
         </md-menu-item>
        </md-menu-content>
-</md-menu>
+    </md-menu>
+<!--
+    ng-if="vm.productConfig['narrowband_cbf_products'] && vm.subarray.band"
+ -->
+    <div ng-if="parent.vm.productsWithNarrowBands.includes(vm.subarray.product) && vm.subarray.band"
+        class='narrow-band-configs'>
+      <md-tooltip>Narrow 1 Frequency in MHz must be between
+        {{parent.vm.getNarrowFreqLimits(1)[0]}}
+        and {{parent.vm.getNarrowFreqLimits(1)[1]}}</md-tooltip>
+      <i>Narrow1 (MHz):</i>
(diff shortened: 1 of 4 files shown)


app/services/obs-sched-service.js
```diff
@@ -549,8 +562,15 @@
                     } else if (sensorName.endsWith('allocations')) {
                         api.sensorValues[sensorName].parsedValue = sensor.value !== "" ? JSON.parse(sensor.value) : [];
                     } else {
-                      if (api.subarrays[subarrayIndex].state != 'inactive')
-                          api.subarrays[subarrayIndex][trimmedSensorName] = sensor.value;
+                      if (api.subarrays[subarrayIndex].state != 'inactive') {
+                        if (trimmedSensorName.match(/requested_narrow._centre_frequency/g)) {
+                          api.subarrays[subarrayIndex][trimmedSensorName]
+                                    = sensor.value/1000000;
```
ajoubertza: Generally we try to keep the internal variables in SI units, and then just scale when displaying.  Maybe that is tricky to do with the editable label, and that why this approach was chosen.

app/scheduler/templates/subarray-config-container.html
```diff
@@ -72,7 +72,24 @@
...
+    </md-menu>
+<!--
+    ng-if="vm.productConfig['narrowband_cbf_products'] && vm.subarray.band"
+ -->
+    <div ng-if="parent.vm.productsWithNarrowBands.includes(vm.subarray.product) && vm.subarray.band"
+        class='narrow-band-configs'>
+      <md-tooltip>Narrow 1 Frequency in MHz must be between
+        {{parent.vm.getNarrowFreqLimits(1)[0]}}
+        and {{parent.vm.getNarrowFreqLimits(1)[1]}}</md-tooltip>
+      <i>Narrow1 (MHz):</i>
```
bngcebetsha: It would seem we are only supporting only 1 narrowband instrument, is that our intention?

PR #518, opened 2019-11-01
Add user log context menu on the star burst GUI page

xinyuwu:
[MT-1037](https://skaafrica.atlassian.net/browse/MT-1037)

Get right menu item to work without using scope() and isolateScope().


xinyuwu:
> On the dev box it still doesn't bring up the context menu and complains about not being able to read whichTooltip function. I've attached the error below
> 
> Screenshot 2019-11-01 at 17 10 18

I have just checked in all the dist files, was going to check in dist files after PR review. dev box should work now.

bxaia:
Thanks I've tested it now on both devb and devs and all works well.

PR #524, opened 2019-12-06
Show remaining time before next global sync

bxaia:
This PR displays the minimum remaining time to the Next Global Sync on the subarray page. The value change is triggered on even t 

Screenshot 2019-12-06 at 16 20 14

Screenshot 2020-01-07 at 15 47 26
Screenshot 2020-01-07 at 15 47 44
Screenshot 2020-01-07 at 15 47 53

JIRA: [MT-1109](https://skaafrica.atlassian.net/browse/MT-1109)


app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -33,6 +33,13 @@
 
                         <span flex></span>
 
+                        <md-menu style="margin: 0 248px; padding: 0; position: relative" class="subarray-mode-button">
+                            <span md-ink-ripple style="margin: 0; padding: 0px" ng-click="$mdMenu.open($event)" title="Remaining Time to Next Global">
+                                <span class="subarray-select-chevron" style="top: 7px"></span>
+                                <span style="margin-left: 16px">NGS: {{ vm.minGlobalSyncTime() ? vm.minGlobalSyncTime() : 'None ' }}<b> Hrs</b></span>
```
bngcebetsha: Just a suggestion - `N.G.S`? Or make the tooltip say `Next Global Sync` instead,...might not be as good as I think though.

app/scheduler/scheduler-home.js
```diff
@@ -1064,13 +1064,23 @@
...
 
                     vm.sensorsRegex = [
                         vm.subarraySensorNames.join('|'),
                         vm.schedSensorNames.join('|'),
                         vm.katpoolSensorNames.join('|'),
                         'state$',
-                        'ridx_position'
+                        'ridx_position',
+                        'dig_l_band_time_remaining',
+                        'dig_u_band_time_remaining'
```
lanceWilliams: How do band X and S get managed? Do we leave them out at this stage?

ppakotze: With 16 S-band installed on site it would be nice to have this visible for S-band as well.  For X-band which is still a long way off it is probably not so useful.

ajoubertza: It would be nice to include S-band, but they don't have a sensor like this.

@bxaia Will the code still run happily if we add the S-band sensors to the GUI code, but the actual sensors are missing?

@ppakotze Is the DMC even doing a global sync for S-band?  We don't verify S-band in the `global_sync.py` script.

app/scheduler/scheduler-home.js
```diff
@@ -1064,13 +1064,23 @@
                         systemConfig['katconn:arrays'].ants,
                         '^(' + systemConfig['katconn:arrays'].ants.replace(/,/g, '|') + ').ridx_position$', true)
                     .then(handleListSensorsResult, handleListSensorsError);
+                    MonitorService.listSensorsHttp(
+                        systemConfig['katconn:arrays'].ants,
+                        '^(' + systemConfig['katconn:arrays'].ants.replace(/,/g, '|') + ').dig_l_band_time_remaining$', true)
+                    .then(handleListSensorsResult, handleListSensorsError);
+                    MonitorService.listSensorsHttp(
+                        systemConfig['katconn:arrays'].ants,
+                        '^(' + systemConfig['katconn:arrays'].ants.replace(/,/g, '|') + ').dig_u_band_time_remaining$', true)
```
ajoubertza: We can reduce this to a single call, since it users a regular expression.  The `listSensorsHttp` request does result in quite a bit of work in the backend, and all GUIs could be requesting this, so better to reduce the number of calls.  We can combine the two new requests with the `ridx_position` request.  It isn't pretty, but you could try this:

```
'^(' + systemConfig['katconn:arrays'].ants.replace(/,/g, '|') + ').(ridx_position|dig_[lsux]_band_time_remaining)$'
```

The `[lsux]` would match any single character from that list, so we would get all bands.
(If we have to avoid bands that don't have the sensors, then remove those letters.  Hopefully we can keep all bands)

app/scheduler/scheduler-home.js
ajoubertza: Similarly, these are also regular expressions, so `dig_[lsux]_band_time_remaining` should work.

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -33,6 +33,13 @@
 
                         <span flex></span>
 
+                        <md-menu style="margin: 0 248px; padding: 0; position: relative" class="subarray-mode-button">
+                            <span md-ink-ripple style="margin: 0; padding: 0px" ng-click="$mdMenu.open($event)" title="Remaining Time to Next Global Sync">
+                                <span class="subarray-select-chevron" style="top: 7px"></span>
+                                <span style="margin-left: 16px">NGS: {{ vm.minGlobalSyncTime() ? vm.minGlobalSyncTime() : 'None ' }}<b> Hrs</b></span>
```
ajoubertza: I see the `Hrs` are still bold.  Will the value also be bold?

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,19 @@
             }
         };
 
+        vm.minGlobalSyncTime = function () {
+            var allReceptors = ConfigService.systemConfig["antenna_labels"]["ALL"].split(',');
```
ajoubertza: The acceptance criteria for this story was to only consider the receptors assigned to the subarray.  There might be some broken digitisers that won't be used, and we don't want to consider their time remaining.

app/scheduler/scheduler-home.js
ppakotze: Yes I believe the global_sync script tries to sync all digitisers marked ready.  Marc might be able to explain better.

app/scheduler/subarray-resources/subarray-resources.html
bngcebetsha: Yes the value will also be bold.

bngcebetsha:
Addressing PR comments, see attached screenshots showing what happens when resizing. Also showing a display from different subarrays.

Screenshot 2019-12-13 at 12 46 35

Screenshot 2019-12-13 at 12 57 04

app/scheduler/scheduler-home.js
bngcebetsha: @ajoubertza the latest commit handles the running of the code without the actual sensors

bngcebetsha:
Screenshot 2019-12-13 at 15 28 38

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
             }
         };
 
+        vm.minGlobalSyncTime = function () {
+
+          // can't work out sync time if initialisation has not finished
+          if (!ConfigService.systemConfig || !vm.subarray)
+            return undefined;
```
ajoubertza: Good to check these kinds of things to make the code more robust.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+
+          var sensorValues = ObsSchedService.sensorValues[vm.subarray.name + '_allocations'];
```
ajoubertza: The variable name isn't so good - it doesn't say which sensor values we will have, and it only gets used 8 lines later.  A better name would help, .e.g. `allocationSensorValues`, or move the usage closer:
```javascript
var sensorValues = ObsSchedService.sensorValues[vm.subarray.name + '_allocations'];
if (!sensorValues)
   return undefined;
var allocations = sensorValues.parsedValue;
```

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+
+          if (!sensorValues)
+            return undefined;
+
+          var allReceptors = ConfigService.systemConfig["antenna_labels"]
```
ajoubertza: This variable is unused.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+                                ["ALL"].split(',');
+          var nextGlobalSyncTime = [];
```
ajoubertza: A plural name would be better for a list, as has been done for the other variables.  `nextGlobalSyncTimes`.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+          var assignedResources = [];
+          var allocations = sensorValues.parsedValue;
+
+          for (var i=0; i<allocations.length; i++) {
+              var receptorName = allocations[i][0];
```
ajoubertza: Note that allocations will contain all resources in the subarray, which include data proxies.

app/scheduler/subarray-resources/subarray-resources.js
ajoubertza: This is OK, but note that the `pool_resources` subarray sensor would be slightly easier to use - it is just a comma separated list of the resources.  The `allocations` sensor includes additional details to indicate which schedule block each resource is allocated to.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+              var nextGlobalSync = undefined;
+
+              if (vm.subarray.band) {
+                var sensor = $scope.parent.vm.sensorValues
+                    [receptorName +"_dig_" + vm.subarray.band + "_band_time_remaining"];
+                if (sensor)
+                  nextGlobalSync = sensor.value;
+              }
+              if (nextGlobalSync) {
```
ajoubertza: If all the receptors have `0` hours remaining will the GUI show the correct result?
What just one antenna has a `0` and the rest still have few hours to go?

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+                nextGlobalSyncTime.push(nextGlobalSync);
```
ajoubertza: The variable names don't make it obvious what is happening here.  Some suggestions:
```javascript
nextGlobalSyncTimes.push(nextGlobalSyncTime);
```
```javascript
nextGlobalSyncTimes.push(receptorNextGlobalSyncTime);
```
```javascript
nextGlobalSyncTimes.push(receptorTimeRemaining);
```

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -72,6 +72,46 @@
...
+              }
+          }
+
+            var minTime = -1;
+            for (var i=0; i<nextGlobalSyncTime.length; i++) {
+              if (minTime < nextGlobalSyncTime[i]) {
+                return nextGlobalSyncTime[i];
```
ajoubertza: This doesn't look like it will work.  It will just return the first value that is > -1.   The old code that used `Math.min.apply` was better - why was that changed?  No need to write a new min function.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -3,7 +3,7 @@
     angular.module('katGui.scheduler')
         .controller('SubArrayResourcesCtrl', SubArrayResourcesCtrl);
 
-    function SubArrayResourcesCtrl($state, $scope, ObsSchedService, $rootScope, $stateParams,
+    function SubArrayResourcesCtrl($state, $scope, ObsSchedService, MonitorService, $rootScope, $stateParams,
```
ajoubertza: MonitorService seems to be unused in this file?

bngcebetsha: yes,..removing it.

app/scheduler/subarray-resources/subarray-resources.js
bngcebetsha: calling `parsedValues` on `ObsSchedService.sensorValues[vm.subarray.name + '_pool_resources']` likes this: `ObsSchedService.sensorValues[vm.subarray.name + '_pool_resources'].parsedValue` returns `undefined`, resulting in empty update of the remaining time sensor value.

While `ObsSchedService.sensorValues[vm.subarray.name + '_allocations'].parsedValue` returns an actual sensor object.

app/scheduler/subarray-resources/subarray-resources.js
bngcebetsha: **Description**

Because `min()` is a static method of Math, you always use it as `Math.min()`, rather than as a method of a Math object you created (Math is not a constructor).

If no arguments are given, the result is `Infinity`.
If at least one of arguments cannot be converted to a number, the result is `NaN`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min

app/scheduler/subarray-resources/subarray-resources.js
ajoubertza: As discussed, we see that not all sensors have `parsedValue` field.  For the `pool_resources` we can just use `value`, which will give the CSV string.

app/scheduler/subarray-resources/subarray-resources.js
ajoubertza: The old code in the PR had something like:  
```javascript
> nextGlobalSyncTimes = [1, -2, 5]
> Math.min.apply(null, nextGlobalSyncTimes)
>> -2
```
Yes, you will need to make sure that that `min` is not used if the list is empty, and that the list doesn't contain items that aren't numbers.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -73,43 +73,33 @@
...
-                nextGlobalSyncTime.push(nextGlobalSync);
+                var sensor = $scope.parent.vm.sensorValues[receptorName +"_dig_" + vm.subarray.band + "_band_time_remaining"];
+                if (sensor) {
+                  var receptorTimeRemaining = sensor.value;
+                  if (minTimeRemaining==undefined) {
+                    minTimeRemaining = receptorTimeRemaining;
+                  }
+                  else if (receptorTimeRemaining && (receptorTimeRemaining<minTimeRemaining)) {
+                    minTimeRemaining = receptorTimeRemaining;
+                  }
```
ajoubertza: It still looks like this will fail if one of the antennas (not the first) has a time of `0` remaining.

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -29,17 +29,14 @@
                                     </md-button>
                                 </md-menu-item>
                             </md-menu-content>
-                        </md-menu>
-                        <md-menu-content style="margin: 0; padding: 0; position: absolute">
                             <md-button class="subarray-config-menu-button" title="Remaining Time to Next Global Sync" ng-click="$mdMenu.open($event); $event.stopPropagation()"
                                 style="font-size: 20px">
                                 <span class="subarray-select-chevron" style="top: 12px"></span>
-                                <span style="margin-left: 14px">
+                                <span style="margin-left: 400px">
```
ajoubertza: Why this change?  The commit message doesn't mention it.

bngcebetsha: before this change, the `NGS` button was under the `subarray_` button, not a good sight.

bngcebetsha:
Gui behaviour after the most recent commit.

No product(band) set yet
Screenshot 2019-12-18 at 12 41 03

One receptor `time_remaining_` set to `0 Hrs` in `l`
Screenshot 2019-12-18 at 12 32 16

One receptor `time_remaining_` set to `22 Hrs` in `l`
Screenshot 2019-12-18 at 12 32 44

`u-band` all set at `48 Hrs`
Screenshot 2019-12-18 at 12 33 10

`s-band` sensor does not exist yet
Screenshot 2019-12-18 at 12 35 13

`x-band` sensor does not exist yet
Screenshot 2019-12-18 at 12 35 24

Resizing the browser
Screenshot 2019-12-18 at 12 43 52

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -29,14 +29,20 @@
                                     </md-button>
                                 </md-menu-item>
                             </md-menu-content>
+                            <md-button class="subarray-config-menu-button" title="Remaining Time to Next Global Sync" ng-click="$mdMenu.open($event); $event.stopPropagation()"
+                                style="font-size: 20px">
+                                <span class="subarray-select-chevron" style="top: 12px"></span>
+                                <span style="margin-left: 400px">
+                                    NGS : <b>{{vm.minGlobalSyncTime()==undefined ? '--- ' : vm.minGlobalSyncTime()}} Hrs</b>
```
ajoubertza: @ofaletse, @xinyuwu  How often will this function get called?  It doesn't seem to be linked to the sensor updates.  Maybe we want to use a variable rather than calling the function every time.

lanceWilliams:
I still get some wrapping and overlapping in resizing which doesn't look too great. I tried some variants using plain md-buttons instead of spans which improves the placement towards the right-hand side of the toolbar but can still cause a nasty overlap on the left-hand side when resizing (the subarray selection dropdown spills out of the toolbar container).

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -29,14 +29,20 @@
                                     </md-button>
                                 </md-menu-item>
                             </md-menu-content>
+                            <md-button class="subarray-config-menu-button" title="Remaining Time to Next Global Sync" ng-click="$mdMenu.open($event); $event.stopPropagation()"
```
lanceWilliams: I think this might be better to position outside of the subarray selection menu container (not sure why it is in here in the first place) since it lets you avoid some of the extra span stuff like on line 35. I tried but was unable to get the full toolbar to resize nicely in all cases, but I think it might generally be better to try and get it to work as just a well defined set of flex objects rather than with interspersed spans. It is definitely tricky as I found out but hopefully the behaviour would be more consistent.

PR #530, opened 2020-01-17
update the parent subarray object

ofaletse:
Subarray object is created when the subarray-resources page is opened. Again subarray object is created when the scheduler-home page is opened. The subarray-resources page call functions like `setProduct()` defined in scheduler-home page. When routing or transitioning to scheduler-resources page it is possible to not open scheduler-home this will imply the subarray object(scheduler-home) will be the old one, as an example when the user set the product from subarray-resources page(config-container) the old subarray object in scheduler-home will be used, and nothing will happen because scheduler-resources page now have a new subarray object. 
In this PR I called `checkCASubarrays` that will ensure that we update the scheduler-home subarray object.

*Screenshots is pointless, maybe gif file*

JIRA: [MT-1185](https://skaafrica.atlassian.net/browse/MT-1185)


ofaletse:
Thanks for the review.

PR #534, opened 2020-02-24
Include CBF boards-marked-assigned sensor

bxaia:
This PR modifies the tooltip for Available skarab boards to include the 'boards_marked_assigned" sensor.

See screenshots below:*

Screenshot 2020-02-24 at 12 05 32

Screenshot 2020-02-24 at 12 12 37

JIRA: [MT-1268](https://skaafrica.atlassian.net/browse/MT-1268)

### DIFF
--- a/app/scheduler/subarray-resources/subarray-resources.html
+++ b/app/scheduler/subarray-resources/subarray-resources.html
@@ -149,10 +149,16 @@
                                 <md-tooltip>Indexer Position</md-tooltip>
                             </span>
 
-                            <span ng-class="{'badge': vm.boardsMarkedUp(resourceAlloc[0])!=undefined}"
+                            <span ng-class="{'badge': (parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value) !=undefined}"
                                 style="background-color: #01717e; position: absolute; left: 100px; top: 4px; font-size: 12px">
-                                {{vm.boardsMarkedUp(resourceAlloc[0])}}
-                                <md-tooltip>{{(vm.boardsMarkedUp(resourceAlloc[0])==undefined ? '---' : vm.boardsMarkedUp(resourceAlloc[0])) + '/' + (vm.totalBoards(resourceAlloc[0])==undefined ? '---' : vm.totalBoards(resourceAlloc[0]))}}  Available SKARAB Boards</md-tooltip>
+                                {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value}}
+                                <md-tooltip class="skarab-boards-tooltip">
+                                    <b><u>Number of SKARAB Boards</u></b><br>
+                                    Available: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value}}<br>
+                                    In Use: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_assigned'].value}}<br>
+                                    Standby: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_standby'].value}}<br>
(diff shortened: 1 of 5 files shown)


app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -149,10 +149,16 @@
                                 <md-tooltip>Indexer Position</md-tooltip>
                             </span>
 
-                            <span ng-class="{'badge': vm.boardsMarkedUp(resourceAlloc[0])!=undefined}"
+                            <span ng-class="{'badge': (parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value) !=undefined}"
```
ajoubertza: Won't this result in lots of exceptions in the console? This  section of HTML isn't specific to CBF resources is it?  E.g., these sensors won't exist for receptor m003.

In addition, even if it is a CBF resource, there was a comment in the previous PR about handling exceptions if the sensor objects don't exist yet on the client side.  It will take some time for the `listSensorsHttp` request to complete.

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -149,10 +149,16 @@
...
-                                {{vm.boardsMarkedUp(resourceAlloc[0])}}
-                                <md-tooltip>{{(vm.boardsMarkedUp(resourceAlloc[0])==undefined ? '---' : vm.boardsMarkedUp(resourceAlloc[0])) + '/' + (vm.totalBoards(resourceAlloc[0])==undefined ? '---' : vm.totalBoards(resourceAlloc[0]))}}  Available SKARAB Boards</md-tooltip>
+                                {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value}}
+                                <md-tooltip class="skarab-boards-tooltip">
+                                    <b><u>Number of SKARAB Boards</u></b><br>
+                                    Available: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_up'].value}}<br>
+                                    In Use: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_assigned'].value}}<br>
+                                    Standby: {{parent.vm.sensorValues[resourceAlloc[0] + '_boards_marked_standby'].value}}<br>
+                                    Total Boards: {{vm.totalBoards(resourceAlloc[0])}}
```
ajoubertza: For consistency, I suggest removing "Boards" from the last line, so "Total" instead of "Total Boards".  The other counts don't include the word "Boards".

app/scheduler/subarray-resources/subarray-resources.html
bngcebetsha: L145 above is another example where a sensor is being called directly instead of wrapping it around a method in JS. We realised that we are implementing 3 methods for the states of the boards, they look exactly the same(clumsy). One option was to make one function that checks all 3 states and return a compact datatype, e.g:

```Javascript
countBoards = function() {
...
return [boardsUp, boardsStandby, boardsAssigned];
}
```
and then call that using indexing on html. However there is indexing here as well which will confuse things.

ajoubertza: OK, so what we see if JavaScript/Angular hide these type of lookup failures away, just returning `undefined`.

bxaia:
> Please make sure commits build without error.
> 
> We see this error: Error 503 Backend is unhealthy

There was an intermittent issues with Github & Jenkins connection, I've re-run the branch and it passes. Please review again

app/scheduler/subarray-resources/subarray-resources.html
ajoubertza: @bxaia or @bngcebetsha Please update the screenshots in the description, so that we have the correct images in the release notes.

PR #537, opened 2020-03-18
Dockerized KATGUI for easier development and debugging

mmphego:
This PR adds the following:
- Containerised KATGUI
- Updated the README file

JIRA: [MT-1323](https://skaafrica.atlassian.net/browse/mt-1323)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+# hadolint ignore=DL3006,DL3007
+FROM docker-registry.camlab.kat.ac.za/cambase_bionic
```
ajoubertza: Wouldn't a basic ubuntu image work too?  Just because this repo is public, may as well use a Docker image that is public too.

tockards: ye, I forgot that this was a public repo. 
We should go for with a basic ubuntu image @mmphego you going to take that on?

mmphego: Oh yeah, thanks for picking that up.

tockards:
@ajoubertza 
There was two issues.
1.) Could not access katgui after exposing the port 8000 on the host machine. 
2.) All js and css files returned with 404 when accessing `http://localhost:8000/katgui` (and this was suspected to be a CORS issue).

We solved point 1 by making the following changes to the `gulp.js` file see https://github.com/ska-sa/katgui/blob/169ffe6ba3e2a0ac7905ba6ef9d8ab90f658c6d9/gulpfile.js#L174

and point 2 was solved by accessing the correct endpoint for local development and the readme notes amended to be helpful.

PR #538, opened 2020-03-19
Handling empty string values when plotting discrete sensor values on sensograph

bxaia:
This PR seeks to fix the sensor graph not plotting correctly the discrete sensor values. The picture below shows continuous plot of the sensor values where empty string values are represented as `(empty)` in the graph. Below are two screenshots of the before and after on same data plotted with two discrete sensors (subarray_1_active_sbs & subarray_1_band) :

Please see **below** the screenshot of the sensor graph before:
Screenshot 2020-03-27 at 17 53 16

Below is the sensor graph image **after** the fix:
Screenshot 2020-03-27 at 17 51 14

JIRA: [MT-1323](https://skaafrica.atlassian.net/browse/MT-1323)


bxaia:
This is now how the graph looks like, displaying the empty spaces where no schedule block was active at the time:
 
Screenshot 2020-03-20 at 15 24 50

app/d3/multi-line-chart.js
```diff
@@ -356,6 +357,7 @@ angular.module('katGui.d3')
                 }
 
                 line = d3.svg.line()
+                    .defined(function(d) { return d.value != ""; })
```
ajoubertza: Would it look any better if we still plotted lines, but replaced the empty strings with something like "<null>" instead of leaving gaps?  Then we'd see steps going to null, and then back to the next value value.

app/d3/multi-line-chart.js
```diff
@@ -58,7 +58,8 @@ angular.module('katGui.d3')
 
             $timeout(function() {
                 scope.unbindResize = scope.$watch(function() {
-                    return element[0].clientHeight - margin.top - margin.bottom + ', ' + element[0].clientWidth;
+                    if (element)
+                        return element[0].clientHeight - margin.top - margin.bottom + ', ' + element[0].clientWidth;
```
ajoubertza: What about the `else` case?  This function will return nothing - is that OK for the caller, or would some other string be more useful.

app/d3/multi-line-chart.js
```diff
@@ -167,6 +168,10 @@ angular.module('katGui.d3')
                             scope.options.yAxisValues = [];
                         }
                         d.value = String(d.value);
+                        if(!d.value) {
```
ajoubertza: Suggest a space is added for consistency
```suggestion
                        if (!d.value) {
```

app/d3/multi-line-chart.js
```diff
@@ -167,6 +168,10 @@ angular.module('katGui.d3')
...
+                            d.empty_value = true;
```
ajoubertza: We were going to use this in the CSV download to go back to the empty string, but not necessary anymore.   CSV files will have the `(empty)` string, matching what we see on the graph.  Line 172 could be removed.

app/d3/multi-line-chart.js
mmphego: The `if` on its own should suffice, we do not envisage it to cause any problems hence why we did not put an else statement.

ajoubertza: OK

PR #542, opened 2020-04-29
Receiver readiness sensor display

bngcebetsha:
We display the 'is receiver ready?' icon, this icon reads the value of the aggregate sensor `agg_m0XX_rsc_rx[band]_ready` that lives on each receptor for each receiver band(l, s, u, x). We are currently implementing the feature for `l` and `u` band as these are the only receivers that exist on site. The sensor reads `true` if all the rules below are satisfied, and false otherwise:

   - `m0XX_rsc_rx[band]_rfe1_temperature: value < 30.0`
   - `m0XX_rsc_rx[band]_lna_h_power_enabled: value == True`
   - `m0XX_rsc_rx[band]_lna_v_power_enabled: value == True`
   - `m0XX_rsc_rx[band]_amp2_h_power_enabled: value == True`
   - `m0XX_rsc_rx[band]_amp2_v_power_enabled: value == True`

The icon on the respective receptor will show `NOMINAL` colour(green) if all rules are satisfied and `WARN` colour(yellow) otherwise. There is a tooltip "Receiver Ready" when you hover over each receptor that is ready, and "Receiver NOT Ready" otherwise.

See screeshots attached ...

We have also modified indexer position to display tooltip description instead of the downward arrow which used to be there but had no other reason to exist other than to hold the tooltip.

When the receiver_ready icon is clicked, it navigates to aggregate sensor-list for the current receptor and band, the user will see which of the four sensors above has been violated the rules if they click the `value` attribute on `sensor-list`.

[MT-1114](https://skaafrica.atlassian.net/browse/MT-1114)

Screenshot 2020-05-04 at 16 05 25

Screenshot 2020-05-04 at 16 05 33

Screenshot 2020-05-04 at 16 07 02

Screenshot 2020-05-04 at 16 07 17
Screenshot 2020-04-29 at 17 48 47

Screenshot 2020-04-30 at 13 34 31

Screenshot 2020-04-30 at 13 36 20

Screenshot 2020-04-30 at 13 36 30

Screenshot 2020-04-29 at 17 47 05

Screenshot 2020-04-29 at 17 59 10


app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -145,6 +145,27 @@
             }
         };
 
+        vm.isReceiverReady = function (receptorName) {
+            // return `true` if the receiver of the currently selected band is ready
```
ofaletse, line 149: The comments might be misleading. To me it looks like we are returning an array with ['IS ready'/'NOT ready', 'monctl'].

bngcebetsha: well spotted Coach, forgot to update the comment

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -149,6 +149,11 @@
                                 <md-tooltip>Indexer Position</md-tooltip>
                             </span>
 
+                            <span class="icon-button fa fa-thumbs-up resource-sensor-status-item" ng-class="vm.sensorClass(parent.vm.sensorValues['mon_' + vm.isReceiverReady(resourceAlloc[0])[1] + '_agg_' + resourceAlloc[0] + '_rsc_rx'+ vm.subarray.band + '_ready'].status)"
```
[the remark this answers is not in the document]
bngcebetsha: we have changed to something a little more professional than a thumbs-up. We are still not sure if a plain square icon will be a good indication, also we are struggling to get a solid yellow to indicate 'warn', we also need to test for when the sensor is 'nominal' for that we need all children sensors to be in status NOMINAL. We could not change them on our simulator.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -145,6 +145,33 @@
...
+                    } else if (sensor.value == false) {
+                        sensor.value = 'NOT ready'
+                    }
+                }
+            }
+            return sensor;
+        }
+
+        vm.navigateToSensorList = function(component, receptorName, filter) {
+            var component = vm.isReceiverReadySensor(receptorName).name.split('_').splice(0, 2).join('_');
```
ajoubertza, line 168: When the string parsing gets complicated like this, it is nice to have a comment with an example of the string being parsed, and showing the expected result (the component in this case)

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -145,6 +145,33 @@
...
+            // return array `sensor`, the full receiver ready sensor object
+            if (!vm.subarray.band)
+                return undefined;
+
+            var subscribedSensors = Object.keys($scope.parent.vm.sensorValues);
+            for (var i=0; i<subscribedSensors.length; i++) {
+                if (subscribedSensors[i].endsWith('ready') && receptorName == subscribedSensors[i].split('_')[3] && subscribedSensors[i].split('_')[5].endsWith(vm.subarray.band)) {
+                    var sensor = $scope.parent.vm.sensorValues[subscribedSensors[i]];
+                    if (sensor.value == true) {
+                        sensor.value = 'IS ready';
```
ajoubertza, line 158: Modifying the sensor value could be problematic, since this sensor object is not owned by the `subarray-resources` controller.  For example, if another part of the GUI is relying on the same sensor, it could break.  I can see that the "IS ready" and "NOT ready" text is useful on the view (HTML) though.

A couple of options (there could be more):
- Leave the value unchanged, and build the text in the view based on true or false value.
- Store this string in a different attribute in the sensor object, that no other part of the GUI is going to look at.  Then use that in the view.

I prefer the first option, since that way all the text for the tooltip is defined in the view.  I.e., a slightly better separation of concerns.  Currently, we have it split between two files: "Received " in the HTML and the "IS ready" coming from JavaScript.

app/scheduler/subarray-resources/subarray-resources.js
```diff
@@ -145,6 +145,33 @@
...
         };
 
+        vm.isReceiverReadySensor = function (receptorName) {
+            // return array `sensor`, the full receiver ready sensor object
+            if (!vm.subarray.band)
+                return undefined;
+
+            var subscribedSensors = Object.keys($scope.parent.vm.sensorValues);
+            for (var i=0; i<subscribedSensors.length; i++) {
+                if (subscribedSensors[i].endsWith('ready') && receptorName == subscribedSensors[i].split('_')[3] && subscribedSensors[i].split('_')[5].endsWith(vm.subarray.band)) {
```
ajoubertza, line 155: Can we please split this over multiple lines - it doesn't fit on the screen.

app/scheduler/subarray-resources/subarray-resources.js
bngcebetsha: you are right, changing value to "NOT ready" instead of `false` is not pretty

bngcebetsha:
> Glad to see that things are working out. Very Good.

I guess github will only allow a merge after you approve Coach

ajoubertza:
@bngcebetsha Can you please add a screen shot showing the receivers in the "ready" state (green).  It will be nice for the release notes to show both ready and not ready.

PR #546, opened 2020-05-07
Matching the color of icon to value of the aggregate sensor

bxaia:
This pull request, changes the _Receiver Ready_ icon colour to match the value of the aggregate sensor instead of using the status of sensor to set the colour. It's implementing 2 colours [Nominal and Warn] colours to match the two sensor values expected.
It also changes the tooltip to reflect the band selected on the receiver

*Below are Screenshots of site with M045 in a warn state and the corresponding sensor list page with the sensor value:*
Screenshot 2020-05-07 at 12 40 17

Screenshot 2020-05-07 at 12 40 49

JIRA: [MT-1449](https://skaafrica.atlassian.net/browse/MT-1449)


app/scheduler/scheduler.less
```diff
@@ -778,3 +778,11 @@
 .skarab-boards-tooltip {
     height: auto;
 }
+
+.receiver-ready {
+    color: #259b24;
```
ajoubertza, line 783: I've very little CSS knowledge, but is it possible to inherit this colour from the `.nominal-sensor-list-item` item instead of hardcoding the colour?  Or just use the existing sensor class.

Also, this colour matches `@alarm-nominal-bg: #259b24;`, not the colour it used to be:
https://github.com/ska-sa/katgui/blob/444ee308f80c4cfd294b9cf813f5aa5d6691dccf/app/sensor-list/sensor-list.less#L70-L72

Why the slight change in the shade of green?

app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -146,11 +146,13 @@
...
+                            <span class="icon-button fa fa-square"
+                                ng-class="{'receiver-ready': vm.isReceiverReadySensor(resourceAlloc[0]).value === true,
+                                'receiver-not-ready': vm.isReceiverReadySensor(resourceAlloc[0]).value === false}"
                                 ng-click="vm.navigateToReceiverReadySensorList(resourceAlloc[0]); $event.stopPropagation()"
                                 ng-if="vm.isReceiverReadySensor(resourceAlloc[0]).value == true || vm.isReceiverReadySensor(resourceAlloc[0]).value == false"
-                                style="position: absolute; left: 70px; bottom: 30px">
-                                <md-tooltip>{{vm.isReceiverReadySensor(resourceAlloc[0]).value == true ? 'Receiver Ready' : 'Receiver Not Ready'}}</md-tooltip>
+                                style="position: absolute; left: 100px; bottom: 25px">
+                                <md-tooltip>{{vm.isReceiverReadySensor(resourceAlloc[0]).value == true ? vm.subarray.band + '-band Receiver Ready' : vm.subarray.band + '-band Receiver Not Ready'}}</md-tooltip>
```
ajoubertza, line 155: Will the band letter be a capital?  L-band is easier to read than l-band.

bxaia: It's current reading as small letter, l make it capital.

app/scheduler/scheduler.less
bxaia: We can customise to inherit the sensor-list-item

bxaia:
Tested all branches with the dist files, below is a screenshot with the x-band, and the icon does not display as expected:

Screenshot 2020-05-11 at 11 13 45

PR #548, opened 2020-05-11
Filter user product  by band 

bxaia:
This pull request implements the band being selected before user product. When the band is  selected, the user products list shows only the selected band-compatible products. When changing bands, the related band-compatible user products are populated, with the default user product selected. 

*Screenshots below:*

**_Band selection enabled_**
Screenshot 2020-05-11 at 14 27 27

**_Band compatible user products populated_**
Screenshot 2020-05-11 at 14 26 15

**_Changing to another band, only those band-compatible user products are displayed_**
Screenshot 2020-05-11 at 14 26 37

JIRA: [MT-1111](https://skaafrica.atlassian.net/browse/MT-1111)


xinyuwu:
Also need to fix build.

app/scheduler/scheduler-home.js
```diff
@@ -306,7 +324,10 @@
 
         vm.setBand = function(band) {
             vm.subarray.band = band;
-            vm.setFrequency(vm.defaultCentreFreqMap[band]);
+            if (band) {
+                vm.setProduct(vm.productsMap[band][0]);
```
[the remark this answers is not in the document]
bngcebetsha: some bands only have 1 product. I will try setting null when band is null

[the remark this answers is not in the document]
ajoubertza: Agree that it is a bit strange to select a "_test" product, but they are only on dev systems, so wouldn't be noticed on site.

What is apparent is that the user products list is sorted in reverse alphabetical order.  I don't know why this was done.  Maybe so that 4k products would show up before 32k?  We could go back to normal order, but that is likely to confuse existing users.

All products on site, in reverse order:
```
In [5]: sorted(kat.katconfig.array_conf.product_conf.keys(), reverse=True)
Out[5]: 
['c875M4k',
 'c875M1k',
 'c856M4k_n54M',
 'c856M4k_n107M',
 'c856M4k4s',
 'c856M4k',
 'c856M32k_n54M',
 'c856M32k_n107M',
 'c856M32k4s',
 'c856M32k',
 'c856M1k',
 'c544M4k',
 'c544M32k',
 'c544M1k',
 'bec856M4kssd',
 'bec856M4kram',
 'bec856M32kssd',
 'bec856M32kram',
 'bec856M1kssd',
 'bec856M1kram',
 'bec544M4kssd',
 'bec544M1kssd',
 'bc856M4k',
 'bc856M32k',
 'bc856M1k',
 'bc544M4k',
 'bc544M32k',
 'bc544M1k']
```

One problem with reverse order is that it will pick the `c856M4k_n54M` product by default for L-band, but this product is hardly ever used.  That's means users will almost always have to change it. 

The utilisation report (1st week in May) shows us that the most common L-band products are c856M4k and bc856M1k.  Can't we add some special cases (hard coded) to pick a better default for each band?

image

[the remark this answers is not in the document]
bngcebetsha: @ajoubertza - please see the latest commit: https://github.com/ska-sa/katgui/pull/548/commits/04e2d95bc356c792cd3b990158988f3fe87293e4

app/scheduler/templates/subarray-config-container.html
```diff
@@ -20,37 +20,37 @@
...
             <md-menu-item ng-if="!parent.vm.dumpRatesMap[vm.subarray.product]">
                 <span>
-                    Select a product first...
+                    Select a band first...
                 </span>
             </md-menu-item>
-            <md-menu-item ng-repeat="band in parent.vm.bandsMap[vm.subarray.product] | orderBy:band track by $index">
-                <md-button ng-click="parent.vm.setBand(band)">
-                    {{band}}
+            <md-menu-item ng-repeat="product in parent.vm.productsMap[vm.subarray.band] | orderBy:product track by $index">
```
ajoubertza, line 51: We still want to keep the reverse sort order.  Old code had a slightly different `orderBy` syntax:
```html
<md-menu-item ng-repeat="product in parent.vm.products | orderBy:'name':true track by $index">
```

app/scheduler/scheduler-home.js
```diff
@@ -129,6 +129,24 @@
...
+                // Notes: This was the best working way to invert a javascript object that we could find.
+                var identity = function(x) {
+                    return x;
+                };
+
+                var reverseMapFromMap = function(map, callback) {
+                    callback = callback || identity;
+                    var reversedMap = {};
+                    _.forEach(map, function(value, key) {
+                    key = callback(key);
```
ajoubertza, line 141: This does nothing, so may as well remove all references to `callback`, and the definition of the `identity` function.

app/scheduler/scheduler-home.js
```diff
@@ -129,6 +129,24 @@
...
+                    reversedMap[value] || (reversedMap[value] = []);
+                    reversedMap[value].push(key);
```
ajoubertza, line 143: As discussed with @bngcebetsha, this reverse mapping won't work correctly for products that are allowed for multiple bands (e.g., c856M4k on RTS).

app/scheduler/scheduler-home.js
```diff
@@ -306,7 +323,21 @@
...
+            } else if(Object.values(vm.productsMap[band]).length == 1) {
+                vm.setProduct(vm.productsMap[band][0]);
+                vm.setFrequency(vm.defaultCentreFreqMap[band]);
+            } else {
+                for (var i=0; i<Object.keys(vm.productsMap[band]).length; i++) {
+                    if(vm.productsMap[band][i].startsWith('c') && vm.productsMap[band][i].endsWith('4k')) {
+                        vm.setProduct(vm.productsMap[band][i]);
+                        vm.setFrequency(vm.defaultCentreFreqMap[band]);
+                    }
+                }
```
ajoubertza, line 339: We should also handle the case of zero products for the current band.  E.g., there are no X-band products on MKAT.

bngcebetsha: Screenshot 2020-05-14 at 17 08 49
Screenshot 2020-05-14 at 17 15 31

PR #555, opened 2020-05-29
Find and close an open in-maintenance user-log for receptors

bngcebetsha:
This pull request automatically closes the user-log that was opened when the(or another) operator took a resource into maintenance. 

- The user-log will be the latest open with compound tag "into_maintenance".
- The user-log for a specified resource  with the compound tag `m0xx_:_into_maintenance` will show as a pop-up dialog. There will be a pre-populated end-time where the user can click "submit" in order to close the log.
- If the current user was not the one who opened the log, they are not able to edit the Content however they can close the log when they click "submit".
- An `m0xx_:_end_maintenance` compound tag and the user's name(with e-mail) will be appended when the resource is being taken out of maintenance.

Screenshot 2020-07-02 at 13 16 15

Screenshot 2020-07-02 at 13 16 32

Screenshot 2020-07-02 at 13 19 29

Screenshot 2020-07-02 at 13 17 01

JIRA: [MT-1498](https://skaafrica.atlassian.net/browse/MT-1498)


app/services/obs-sched-service.js
```diff
@@ -103,6 +103,32 @@
...
+                var query = "?";
+                UserLogService.queryUserLogs(query).then(function (result) {
+                    if (result.data) {
+                        result.data.forEach(function (userlog) {
+                            reportUserlogs.push(UserLogService.populateUserlogTagsFromMap(userlog));
+                        });
+                        var startTimes = [];
+                        for (var i=0; i<reportUserlogs.length; i++){
+                            if (reportUserlogs[i].compound_tags.length > 0) {
+                                var selectedReceptor = reportUserlogs[i].compound_tags[0].split('_:_')[0];
```
xinyuwu, line 119: Can you confirm the '_into_maintenance' will always be the first compound tag?

app/services/obs-sched-service.js
```diff
@@ -103,6 +103,32 @@
...
+                                if (resource == selectedReceptor) {
+                                    startTimes.push(new Date(reportUserlogs[i].start_time).getTime());
+                                    if ((new Date(reportUserlogs[i].start_time).getTime() == Math.max.apply(null, startTimes))) {
```
xinyuwu, line 122: Instead of pushing start_time into an array and find the max, you can alway just remember the current max and compare it to your start time.

app/services/obs-sched-service.js
bngcebetsha: This method creates only this compound tag and our filtering filters only it. Could you please suggest a more robust way to confirm this?

app/services/obs-sched-service.js
```diff
@@ -113,13 +113,15 @@
                         result.data.forEach(function (userlog) {
                             reportUserlogs.push(UserLogService.populateUserlogTagsFromMap(userlog));
                         });
-                        var startTimes = [];
                         for (var i=0; i<reportUserlogs.length; i++){
+                            var latestTime = new Date(reportUserlogs[0].start_time).getTime() || null;
```
ajoubertza, line 117: This doesn't depend on the loop index, `i`, so it doesn't need to be inside the loop.
We could just do `var latestTime = null;` before the loop starts.

app/services/obs-sched-service.js
```diff
@@ -103,6 +103,34 @@
...
+                            if (reportUserlogs[i].compound_tags.length > 0) {
+                                currentTime = new Date(reportUserlogs[i].start_time).getTime();
+                                latestTime = Math.max(latestTime, currentTime);
+                                var selectedReceptor = reportUserlogs[i].compound_tags[0].split('_:_')[0];
+                                if (resource == selectedReceptor) {
+                                    if (new Date(reportUserlogs[i].start_time).getTime() == latestTime) {
+                                        var usersName = reportUserlogs[i].user.name;
+                                        reportUserlogs[i].content +=  '\n' + usersName + ' has taken ' + resource + ' out of maintenance.';
+                                        UserLogService.editUserLog(reportUserlogs[i], true);
+                                    }
```
ajoubertza, line 128: This section is a problem here because we haven't necessarily found the maximum time yet - we are still in the loop searching for `latestTime`.  If the userlogs come back in chronological order, then we'll every user log we look at will be the latest.  If the userlogs come back in a random order, then we may still update a few of them before we hit the one that is actually the latest.

We need to break this into two steps:
1. Find user log with latest start time (e.g., using a loop)
2. Update the user log that we know is the latest one (after the loop).

ajoubertza:
> to automatically and close the user-log that was opened when the(or another) operator took that antenna to maintenance.

My understanding was that only the user that created the user log can edit it.  The sentence above makes is sound like any operator can close the user log opened by another user.
1.  Does the implementation in this PR allow any user to close it?
2.  If yes, which username is associated with the closed user log?

app/services/obs-sched-service.js
```diff
@@ -103,6 +103,34 @@
                     compound_tags: compoundTag,
                 };
                 UserLogService.editUserLog(newlog, true);
+            } else if (maintenance === 'clear') {
+                $state.go('userlogs-report', {
+                    filter: resource + "_:_into_maintenance"})
```
lanceWilliams, line 108: Isn't the filter applied on query results in the GUI? I thought you would use the compound tag in the query that gets used in the service argument on line 111. You should end up with a much smaller set of userlogs to work with.

bngcebetsha:
> > to automatically and close the user-log that was opened when the(or another) operator took that antenna to maintenance.
> 
> My understanding was that only the user that created the user log can edit it. The sentence above makes is sound like any operator can close the user log opened by another user.
> 
> 1. Does the implementation in this PR allow any user to close it?
> 2. If yes, which username is associated with the closed user log?

We have preserved the current functionality, a user can only edit a user log they have opened. Other users can only add extra tags and submit the log. Our new implementation also populates the `end_time` to be the current time of the log closure.

app/userlogs/userlogdialog.tmpl.html
```diff
@@ -128,8 +128,8 @@
         <span ng-show="uploadingFiles" style="font-weight: bold">
             Uploading Files, please wait...
         </span>
-        <md-button ng-click="hide()">{{!editMode && openedWithoutEndTime && openedWithoutStartTime && !chipHasBeenAdded? 'Close' : 'Cancel'}}</md-button>
-        <md-button ng-hide="!editMode && openedWithoutEndTime && openedWithoutStartTime && !chipHasBeenAdded" ng-click="submit()" ng-disabled="endDateTimeError || uploadingFiles || content.length == 0 || !validStartTime || !validEndTime || !content || !validTags">Submit</md-button>
+        <md-button ng-click="hide()">{{!editMode && openedWithoutEndTime && openedWithoutStartTime && chipHasBeenAdded? 'Close' : 'Cancel'}}</md-button>
+        <md-button ng-hide="!editMode && openedWithoutEndTime && openedWithoutStartTime && chipHasBeenAdded" ng-click="submit()" ng-disabled="endDateTimeError || uploadingFiles || content.length == 0 || !validStartTime || !validEndTime || !content || !validTags">Submit</md-button>
```
ajoubertza, line 132: I'm confused.  I don't understand why `!chipHasBeenAdded` changed to `chipHasBeenAdded`.  I know we wanted the submit and cancel buttons to appear when we bring up the userlog to be closed, but in that case the chips (used for simple tags) would not have changed.  Just the end time, content and compound tags would have changed.

Also if a `chipHasBeenAdded`, then we don't want to hide the "Submit" button, so the original logic was correct.  Maybe we need an additional flag for the new case?

app/services/obs-sched-service.js
```diff
@@ -114,22 +115,28 @@
...
-                        logToClose.end_time = moment(endTime.getTime()).format('YYYY-MM-DD HH:mm:ss');
-                        UserLogService.editUserLog(logToClose, logToClose.user.email === $rootScope.currentUser.email, 'userlogDialogContentElement');
+                        if (logToClose) {
+                            var endTime = new Date(new Date().toUTCString().slice(0, -4));
+                            logToClose.end_time = moment(endTime.getTime()).format('YYYY-MM-DD HH:mm:ss');
+                            logToClose.content +=  '\n\nResource ' + resource + ' taken out of maintenance by ' + $rootScope.currentUser.name + ' <' + $rootScope.currentUser.email + '>.';
+                            logToClose.compound_tags.push(compoundTagClear);
+                            UserLogService.editUserLog(logToClose, logToClose.user.email === $rootScope.currentUser.email, 'userlogDialogContentElement');
+                        } else {
+                            NotifyService.showPreDialog('Error finding userlog to close', logToClose);
```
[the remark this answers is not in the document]
bngcebetsha: How about:
`NotifyService.showPreDialog('Error finding userlog to close', "No maintenance log was submitted");`

which will result in what is on the screenshot

Screenshot 2020-06-09 at 15 38 22

app/userlogs/userlogdialog.tmpl.html
bngcebetsha: `chipHasBeenAdded` checks if a new simple tag has been added, the `submit` button is hidden if no new tag is added, we have applied a negate to overwrite this.

bxaia:
Tested this with Lance yesterday and resolved that the errors we experienced was a cache issue, as we were testing on the same tab interchanging the users. Tested it again with Buntu this morning on different browsers with different users and implementation works as expected i.e. User who did not create userlog cannot edit userlog contents and is able to just click Submit with no interaction with the Userlog dialog, and resource is taken out of maintenance

ajoubertza:
Please update the pull request description and screenshots.  We should mention the new compound tag and show it in the screenshots.  Also don't need to add a simple tag anymore, so the images showing the "shift" tag are not necessary.

app/services/obs-sched-service.js
```diff
@@ -103,6 +103,42 @@
...
+                        });
+
+                        var latestTime = null;
+                        var currentTime = null;
+                        var logToClose = null;
+                        for (var i=0; i<reportUserlogs.length; i++) {
+                            for (var j=0; j<reportUserlogs[i].compound_tags.length; j++) {
+                                var selectedResource = reportUserlogs[i].compound_tags[j].split('_:_')[0];
+                                currentTime = new Date(reportUserlogs[i].start_time).getTime();
+                                latestTime = Math.max(latestTime, currentTime);
```
ajoubertza, line 123: It is unlikely, but there may be an edge case where this causes a problem.  It would be safer if these two lines were inside the `if (resource == selectedResource && !reportUserlogs[i].end_time)` block.  We want to close the latest open user log (i.e., no end time set), so we should only update `latestTime` if we find an open user log.

app/services/userlog-service.js
```diff
@@ -383,6 +396,7 @@
                         $scope.openedWithoutEndTime = $scope.end_time !== null && $scope.end_time.length > 0;
                         $scope.chipHasBeenAdded = false;
                         $scope.focusTarget = focusTarget? focusTarget: 'userlogDialogStartTimeElement';
+                        $scope.internalModification = log.modified !== null;
```
ajoubertza, line 399: I don't see any code in the katgui codebase that ever sets `log.modified`, so won't it just be `undefined`?  Thus the check `log.modified !== null` always returns true.  I suspect the "Submit" button would now appear in cases where it didn't use to appear, and shouldn't.

bngcebetsha: you are right,..thanks

app/services/userlog-service.js
```diff
@@ -383,6 +396,7 @@
                         $scope.openedWithoutEndTime = $scope.end_time !== null && $scope.end_time.length > 0;
                         $scope.chipHasBeenAdded = false;
                         $scope.focusTarget = focusTarget? focusTarget: 'userlogDialogStartTimeElement';
+                        $scope.internalModification = log.modified !== undefined;
```
ajoubertza, line 399: Where is the code that will set the `modified` attribute on userlog objects?

PR #556, opened 2020-06-15
Make receiver ready widget show more information

bngcebetsha:
We make the `receiver` sensor display WARN == "yellow" if status is not nominal

JIRA: [MT-1533](https://skaafrica.atlassian.net/browse/MT-1533)


app/scheduler/subarray-resources/subarray-resources.html
```diff
@@ -147,8 +147,8 @@
                             </span>
 
                             <span class="icon-button fa fa-square"
-                                ng-class="{'nominal-sensor-list-item': vm.isReceiverReadySensor(resourceAlloc[0]).value === true,
-                                'warn-sensor-list-item': vm.isReceiverReadySensor(resourceAlloc[0]).value === false}"
+                                ng-class="{'nominal-sensor-list-item': vm.isReceiverReadySensor(resourceAlloc[0]).value == true,
+                                'warn-sensor-list-item': vm.isReceiverReadySensor(resourceAlloc[0]).value == false || vm.isReceiverReadySensor(resourceAlloc[0]).status != 'nominal'}"
```
ajoubertza, line 151: Do we need to check the status in the nominal colour case as well?

E.g., If `value` is true, but `status` is `'error'`, will this result in
`ng-class="{'nominal-sensor-list-item', 'warn-sensor-list-item'}?  In that case, what colour will the button be?

Also, can we split these very long lines?  They don't fit in the unified diff view on github.

bngcebetsha: if `value = false` OR `status != 'nominal' display yellow, else green

ajoubertza: OK, but does it make sense to list both classes in the case that we know it should be a warning?

bngcebetsha: Yes

`ng-class` is accepting a `dict()-type` with key-value pairs, it only displays the value that valuates to `true` corresponding to a key:
`d = {'nominal-sensor-list-item': true_value,`
`'warn-sensor-list-item': false_value || status != 'nominal'}"`
 Our keys are: 
- `'nominal-sensor-list-item'` and
- `'warn-sensor-list-item'`
We need them both...

ajoubertza: As discussed - we don't really want both of them.  Either nominal or warn, not both.  The colour from the last one "wins", so it does end up looking correct.  I.e., yellow when there is a warning.

PR #559, opened 2020-08-17
Download proper zero values on sensor-graph csv download

bngcebetsha:
#### Download proper values for zero
Make the "download_csv" functionality of sensor-graph append proper values for `0` when there is no "Resampling_Interval"

Problem: No Sampling Interval is set, the sensor that has the sampling interval value equates to 0 as a result the opertion that uses this value will result to undefined.

See [MT-1675](https://skaafrica.atlassian.net/browse/MT-1675)


app/d3/multi-line-chart.js
```diff
@@ -909,7 +909,11 @@ angular.module('katGui.d3')
                         } else if (hasStatus) {
                             dataString += ',';
                         }
-                        dataString += (sensorValues.values[i].value || sensorValues.values[i].avg_value);
+						if (sensorValues.values[i].value !== undefined) {
+                            dataString += sensorValues.values[i].value;
+						} else if (sensorValues.values[i].avg_value !== undefined) {
+							dataString += sensorValues.values[i].avg_value;
+						}
```
ajoubertza, line 916: I know it isn't critical in JavaScript, but would still be nice if the indentation was kept consistent.   This whole block should be indented a bit.  And lines 913 and 915 should be the same.

What happens if both `value` and `avg_value` are undefined?  Is it different to what the old code used to do?

ajoubertza, line 916: We could also use `sensorInfo` instead of `sensorValues.values[i]`.  Most of the body of the for loop is doing that.   The code from line 907 could be updated.

bngcebetsha: - My editor shows indentation is ok even though here it isn't...can @amakhaba please fix this
- Lines 913 and 915 can't be the same...
Screenshot 2020-08-18 at 14 25 47
Screenshot 2020-08-18 at 14 26 43

- Will think of a way to handle the case where both `value` and `avg_value` are `undefined`

amakhaba: Will look into it @bngcebetsha

ajoubertza: Probably a mixture of tabs and spaces in the file.

ajoubertza: > Lines 913 and 915 can't be the same...
Sorry, I meant the indentation should be the same, not the code.

PR #562, opened 2020-08-24
Fixing the narrow centre frequency display

bxaia:
This is a quick fix to the narrow band centre frequency display, on the subarray page. 

_**A full screen showing the whole number in the narrow centre frequency**_
Screenshot 2020-08-24 at 09 52 47

_**A minimized screen showing the whole number in the narrow centre frequency**_
Screenshot 2020-08-24 at 09 53 02

### DIFF
--- a/app/scheduler/templates/subarray-config-container.html
+++ b/app/scheduler/templates/subarray-config-container.html
@@ -86,7 +86,8 @@
         ng-class="{'invalid-narrow-freq': !parent.vm.isNarrowBandFreqValid(1)}"
         class="fade-in narrow-freq-input-box ng-pristine ng-untouched ng-empty"
         ng-disabled="!$root.expertOrLO"
-        ng-model="vm.subarray.requested_narrow1_centre_frequency">
+        ng-model="vm.subarray.requested_narrow1_centre_frequency"
+        type="number">
     </div>
 
     <md-menu>
--- a/app/scheduler/templates/subarray-config-container.html
+++ b/app/scheduler/templates/subarray-config-container.html
@@ -86,8 +86,7 @@
         ng-class="{'invalid-narrow-freq': !parent.vm.isNarrowBandFreqValid(1)}"
         class="fade-in narrow-freq-input-box ng-pristine ng-untouched ng-empty"
         ng-disabled="!$root.expertOrLO"
-        ng-model="vm.subarray.requested_narrow1_centre_frequency"
-        type="number">
+        ng-model="vm.subarray.requested_narrow1_centre_frequency">
     </div>
 
     <md-menu>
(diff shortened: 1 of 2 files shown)


app/scheduler/templates/subarray-config-container.html
```diff
@@ -86,8 +86,7 @@
         ng-class="{'invalid-narrow-freq': !parent.vm.isNarrowBandFreqValid(1)}"
         class="fade-in narrow-freq-input-box ng-pristine ng-untouched ng-empty"
         ng-disabled="!$root.expertOrLO"
-        ng-model="vm.subarray.requested_narrow1_centre_frequency"
-        type="number">
```
ajoubertza, line 90: 1.  Does the `type` field also provide some validation?  What will happen now if the user types in something that is not a number?
2. Has this been tested on a dev system?  I.e., activating a subarray with a narrow band frequency specified and verifying the corresponding subarray sensor has the correct value.
3. Does the display work OK with more digits after the decimal point?

ajoubertza (commented):
Thanks for quick fix.  Some questions though.

app/scheduler/templates/subarray-config-container.html
bxaia: 1. The fix has been tested on both devb and site, the subarray activates correctly with the specified narrow band frequency and the correspinding sensor value displays the requested frequency (see attached image with sensor value).
2. I just tested with more digits after the decimal and the current fix only show up to the 1st decimal point, I will then have to increase the field and I think with the increase we might not even have to take out the type as more digits will be accommodated.

ajoubertza: The CBF reports the narrowband frequency in hertz, so worst-case we could get up to 6 digits after the decimal place.

I also notice on devb that the up-down editing controls show up and are clickable even when the subarray is active.  Can we prevent editing if subarray not inactive?
image

bxaia: Alright am looking into disabling editing controls on an active subarray

ajoubertza: Not just _active_, also _initialising_ and _error_.   Only allowed editing in the _inactive_ state.

bngcebetsha (approved):
Looks good to me

ajoubertza (approved):
Seeing as update is urgent, and functionality is improved, we can go with this as is.
Please make a v27 hotfix for deployment.

ajoubertza:
@bxaia Please update the dist files.

bxaia:
> @bxaia Please update the dist files.

Thanks Anton I realised while doing the v27 branch that I had not updated the dist files I've currently restored the branch so I can push to the same branch the dist files

PR #576, opened 2020-10-07
Add proposal id column and number of antennas to the utilization report

amakhaba:
This Pull request adds a `proposal_id` column between the `SB ID` code and `Owner`, as well as the number of antennas allocated to the SB as the last column of the utilization report.

Attached below is a screenshot of the new implementation as well as the exported PDF file of the utilization report.

utilization_report

PDF example:
https://github.com/ska-sa/katgui/files/5340672/utilisation_report_2020-10-07.12_36_14.pdf

JIRA: [MT-1851](https://skaafrica.atlassian.net/browse/MT-1851)


app/reports/utilisation-report.js
```diff
@@ -540,6 +545,7 @@
                             var duration = moment.duration(sb.durationSeconds, 's');
                             sb.duration = vm.durationToString(duration);
                             sb.percentageOfTotal = vm.percentageOfTotalToString(sb.durationSeconds);
+                            sb.n_ants = sb.antennas_alloc.split(",").length;
```
ajoubertza, line 548: If `antennas_alloc` is an empty string, this returns 1.  It is extremely unlikely to occur, but easy to handle, so suggest we handle that special case.

app/reports/utilisation-report.js
```diff
@@ -275,13 +277,16 @@
                     margin: {top: 8, bottom: 8},
                     columnStyles: {
                         id_code: {columnWidth: 85},
+                        proposal_id: {columnWidth: 85},
```
ajoubertza, line 280: In attached example PDF, the proposal IDs are truncated - this width needs to be increased and/or add overflow.  Not sure what the longest proposal ID is.

ajoubertza (changes requested):
Looks very nice.  A few minor notes.

mmphego (approved):
Looks good to me :partying_face:

ajoubertza (approved):
LGTM - thanks for updates.

PR #581, opened 2020-10-12
Display total duration for active SBs

bngcebetsha:
We add all durations for the schedule blocks that were active over a given time window as a summary before the SB details section, see attached screenshots for demo.

#### Gui
Screenshot 2020-10-12 at 11 37 31

#### PDF download
Screenshot 2020-10-12 at 11 37 56

JIRA: [MT-1852](https://skaafrica.atlassian.net/browse/MT-1852)


app/reports/utilisation-report.js
```diff
@@ -254,7 +255,7 @@
                     theme: 'striped',
                     margin: {top: 8, bottom: 8}});
 
-                var sbColumns = [
+                    var sbColumns = [
```
mmphego, line 258: Should this be indented?

app/reports/utilisation-report.js
```diff
@@ -537,13 +562,17 @@
...
                     vm.SBDetails.forEach(function (sb) {
                         if (sb.actual_end_time && sb.actual_start_time) {
                             var startSeconds = moment(sb.actual_start_time, MOMENT_DATETIME_FORMAT).unix();
                             var endSeconds = moment(sb.actual_end_time, MOMENT_DATETIME_FORMAT).unix();
                             sb.durationSeconds = Math.abs(endSeconds - startSeconds);
                             var duration = moment.duration(sb.durationSeconds, 's');
                             sb.duration = vm.durationToString(duration);
+                            totDur += duration;
+                            var totalDur = moment.duration(totDur);
+                            vm.totalDuration = vm.durationToString(totalDur);
```
mmphego, line 575:
```suggestion
                            vm.totalDuration = vm.durationToString(moment.duration(totDur));
```

ajoubertza: Yes.

It would also be better to move the assignment of `vm.totalDuration` outside the loop - there is no need to write to it every time.

Then we wouldn't need the `totalDuration` variable inside the loop.  It would be nice to give `totDur` a better name as well - the name is a bit too short.

app/reports/utilisation-report.js
```diff
@@ -31,6 +31,7 @@
             vm.subarrayProductDurations = {};
             vm.subarrayMaintenanceDurations = {};
             vm.resourceItemColumns = [];
+            vm.totalDuration = 0;
```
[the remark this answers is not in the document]
ajoubertza: Looking at the assignment of the usage of this later, it is set to a string provided by `durationToString`.  Suggest that the initial value matches that type.   So "0", or "00:00", or something like that.

ajoubertza (commented):
Looks nice.  A few suggested improvements.

ajoubertza (approved):
LGTM

mmphego (approved):
Looks good to me, don't forget the dists

PR #583, opened 2021-02-09
Adjust D3 CSS to be compatible with Chrome v88+ to fix plotting

bxaia:
This PR removes the class that is overriding the style in the new Chrome version (88), that prevented the proper plotting of graphs in the GUI.

*Please see below screenshots of fixed displays:*
Screenshot 2021-02-09 at 12 00 15
Screenshot 2021-02-09 at 11 59 53
Screenshot 2021-02-09 at 11 59 07

The change in Chromium that caused the old CSS which was previously ignored to now be interpreted differently is [here](https://chromium.googlesource.com/chromium/src/+/82e2bc96d1193d635240f7f5f3907d8092be440f).

JIRA: [OPS-1632](https://skaafrica.atlassian.net/browse/OPS-1632)


ajoubertza (approved):
LGTM

ofaletse:
Wow.. I am happy to see that you came up with a fix. Thanks very much
I will not mind to get more details to understand the issue in details. 
Also I would prefer if you comment out the code and write the reason why the css is commented out. This has been done previously in the code. I guess it also enlighten others in the future.

bxaia:
Thanks OJ, I will try and elaborate more on the release notes. When I looked at the removed code snippet no where else was it called and therefore opted for removing it rather than commenting it out.

ofaletse:
@bxaia I actually need more technical explanation. Sprint retro will do.
Thanks @ajoubertza

ajoubertza:
> Also I would prefer if you comment out the code and write the reason why the css is commented out. This has been done previously in the code. I guess it also enlighten others in the future.

Sorry, @ofaletse, but I disagree with that approach.  It results in unnecessary clutter.  If it isn't used it must be removed.  Git keeps the history.  This particular bit of CSS has been ignored by browsers for a long time.

ofaletse:
@ajoubertza thanks for your comment.
I hear you, If you look in to the css code you would note that Francois used to add comments to state why he stopped using a specific property. In my view (which I still hold) this is very helpful and easily accessible. You don’t have to go with my suggestion.

PR #591, opened 2021-06-29
Add missing import to download spreadsheet on utilisation report

bxaia:
This PR adds a missing import to fix downloading the utilisation report in a spreadsheet format. 

Screenshot 2021-06-29 at 10 35 05

JIRA: [OPS-2052](https://skaafrica.atlassian.net/browse/OPS-2052)


lanceWilliams (approved):
Looks ok, although I don't understand how this arose. Did the hosting for the xlsx module change that this new source is required?

bxaia:
> Looks ok, although I don't understand how this arose. Did the hosting for the xlsx module change that this new source is required?

I suspect something changed in the xlsx module, as I was looking through the problem it kept complaining about XLSX not being defined.

shoosen (approved):

ajoubertza:
I am also surprised.  I thought the minified katgui build included all the packages required, so we would not be susceptible to changes like this.  Did we miss a step when adding the xlsx package, so that it doesn't end up in the build?

app/reports/utilisation-report.html
```diff
@@ -1,3 +1,4 @@
+<script lang="javascript" src="https://cdnjs.cloudflare.com/ajax/libs/xlsx/0.15.2/xlsx.full.min.js"></script>
```
mmphego, line 1: We failed to pin the version, here: https://github.com/ska-sa/katgui/blob/3ff946374e3043720645992e560ff6f1284d43e0/index.html#L50
This might be the cause!

Perhaps, make the changes in https://github.com/ska-sa/katgui/blob/3ff946374e3043720645992e560ff6f1284d43e0/index.html#L50 instead of here (utilisation report)!

mmphego (changes requested):
Oh well! Seems like PR was merged

app/reports/utilisation-report.html
ajoubertza: @mmphego Nothing else is pinned there.  Isn't the yarn.lock file for pinning?  And package.json?  I see those two json files have `"xlsx": "^0.16.9`, while this PR uses version 0.15.2?

I don't think this PR is the correct solution - it doesn't follow the pattern used for other modules.

bxaia: I'll re-open the branch and make the change on index.html file

PR #592, opened 2021-06-29
Update angular dependency packages 

bxaia:
This PR is updating angular package dependencies as the current version is not downloading the spreadsheet of the utilisation report.

Screenshot 2021-06-29 at 12 22 48

JIRA: [OPS-2052](https://skaafrica.atlassian.net/browse/OPS-2052)


mmphego (approved):
Looks good to me!

ajoubertza:
@bxaia Shouldn't we also remove the explicit link added in PR #591?
https://github.com/ska-sa/katgui/blob/421bf4ff4d0d6ba8a6152273277039e7b2e4d33d/app/reports/utilisation-report.html#L1

Can you confirm that the minified build includes the xlsx javascript, so that we don't need external Internet access for the katgui to work?

bxaia:
> @bxaia Shouldn't we also remove the explicit link added in PR #591?
> https://github.com/ska-sa/katgui/blob/421bf4ff4d0d6ba8a6152273277039e7b2e4d33d/app/reports/utilisation-report.html#L1
> 
> Can you confirm that the minified build includes the xlsx javascript, so that we don't need external Internet access for the katgui to work?

you are right we should remove that.

bngcebetsha (approved):
not  sure how to solve the Jenkins issue

yarn.lock
```diff
@@ -5447,9 +5447,9 @@ wrappy@1:
   resolved "https://registry.yarnpkg.com/wrappy/-/wrappy-1.0.2.tgz#b5243d8f3ec1aa35f1364605bc0d1036e30ab69f"
   integrity sha1-tSQ9jz7BqjXxNkYFvA0QNuMKtp8=
 
-xlsx@^0.16.9:
-  version "0.16.9"
-  resolved "https://registry.yarnpkg.com/xlsx/-/xlsx-0.16.9.tgz#dacd5bb46bda6dd3743940c9c3dc1e2171826256"
+xlsx@^0.15.2:
+  version "0.15.2"
+  resolved "https://registry.yarnpkg.com/xlsx/-/xlsx-0.15.2.tgz#dacd5bb46bda6dd3743940c9c3dc1e2171826256"
   integrity sha512-gxi1I3EasYvgCX1vN9pGyq920Ron4NO8PNfhuoA3Hpq6Y8f0ECXiy4OLrK4QZBnj1jx3QD+8Fq5YZ/3mPZ5iXw==
```
ajoubertza, line 5453: The integrity line, and probably the hash after the version would need to change, if the version is changing.  That's likely why CI is failing:
```
12:46:58  error https://registry.yarnpkg.com/xlsx/-/xlsx-0.15.2.tgz: Integrity check failed for "xlsx" (computed integrity doesn't match our records, got "sha512-0YcFaYgoqSHh9JcWeQU5X/kR6hyXrs+boc6BomPnJAUVYGgo+Jna9blIwMztjQYfJ6WHs5tKtg5GBBGjkLFVhA== sha1-GTjnEuYoEwNbDsmAOhL/qkAM408=")
12:46:58  info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

ajoubertza (dismissed):

ajoubertza:
Now that angular is pinned with `^1.6.10`, will this part of the `yarn.lock` file be OK?
https://github.com/ska-sa/katgui/blob/551e112f00f1bf1bbffa8318cd412cf9faf30196/yarn.lock#L184-L187

yarn.lock
bxaia: After investigation with the team, we found out that xlsx requires an angular version that is >1.4 and <1.7, and the other changes that were implemented earlier.

amakhaba (approved):
LGTM

ajoubertza:
In the CI output, is this warning something we need to be concerned about:
```
[2021-06-29T13:16:53.712Z] warning " > angular-material@1.1.6" has incorrect peer dependency "angular@>=1.4 <1.7".
```
That seems to contradict the commit message for 551e112.

(Although the message was in previous builds, at least a month ago)

yarn.lock
```diff
@@ -186,6 +186,11 @@ angular@^1.0.8, angular@^1.6.3:
   resolved "https://registry.yarnpkg.com/angular/-/angular-1.7.8.tgz#b77ede272ce1b261e3be30c1451a0b346905a3c9"
   integrity sha512-wtef/y4COxM7ZVhddd7JtAAhyYObq9YXKar9tsW7558BImeVYteJiTxCKeJOL45lJ/+7B4wrAC49j8gTFYEthg==
 
+angular@^1.6.10:
+  version "1.8.2"
+  resolved "https://registry.yarnpkg.com/angular/-/angular-1.8.2.tgz#5983bbb5a9fa63e213cb7749199e0d352de3a2f1"
+  integrity sha512-IauMOej2xEe7/7Ennahkbb5qd/HFADiNuLSESz9Q27inmi32zB0lnAsFeLEWcox3Gd1F6YhNd1CP7/9IukJ0Gw==
+
```
ajoubertza, line 193: How does it work if we have multiple entries for the same package?  I.e., which one gets used - is it 1.7.8 or 1.8.2, or some combination?

amakhaba: This should be rectified by the last commit, when we installed the earlier version it was a fix for the xlsx dependency but we later on discovered that the angular-material package was the one with dependency issues, Current version fixes the xlsx installation as well as the angular-material dependency warning

yarn.lock
```diff
@@ -186,6 +186,11 @@ angular@^1.0.8, angular@^1.6.3:
   resolved "https://registry.yarnpkg.com/angular/-/angular-1.7.8.tgz#b77ede272ce1b261e3be30c1451a0b346905a3c9"
   integrity sha512-wtef/y4COxM7ZVhddd7JtAAhyYObq9YXKar9tsW7558BImeVYteJiTxCKeJOL45lJ/+7B4wrAC49j8gTFYEthg==
 
+angular@^1.6.10:
+  version "1.8.2"
```
mmphego, line 190: @amakhaba please re-run that exercise, we omitted the pinning of `angular`. 
Note: Version mismatch as highlighted by @ajoubertza

amakhaba: Where is the mismatch, as far as i am aware is that the currently installed version of angular is 1.8.2 is there any other that exist in this file ?

amakhaba: I saw the mismatch thanks

amakhaba:
Response we got from @rohanschwartz for the package mismatch issue was "So the reason for multiple packages in the lock file with different versions is probably due to the fact that not all your packages in the angular.json file that require the angular package are on the same version that requires the same angular version. I think the package manager handles this fine. ". And also if you want to investigate how many times a depenedency is installed and what packages dependens on it (recursivly), run: `yarn why angular`. This will give you a visual output of what which package depend on which version of angular: 
angular-dependencies.

mmphego (approved):

ajoubertza:
Thanks for the explanation, @amakhaba. Can we please update the PR title and description before merging?  We are no longer pinning xlsx.

PR #596, opened 2021-08-18
Reduce scrolling on sensor list page (and revert to older packages)

bxaia:
This PR reduces scrolling for the components and proxies on the sensor list by increasing the number of columns from 2 to 4, and adjusting the minimum heights. 

**Before:**

image

**After** (small monitor)
image

**After**  (larger monitor):
image

JIRA: [MT-2553](https://skaafrica.atlassian.net/browse/MT-2553)


mmphego (commented):
Does it adapt to the number of proxies/components available(dynamic)? or is it statically set?

amakhaba:
> Does it adapt to the number of proxies/components available(dynamic)? or is it statically set?

Good Question mpho, I want to think it should since our boxes have different proxies and components

ajoubertza:
@bxaia Nice.  What does this look like on a smaller screen?  I have  2880 × 1800.  The screen shot is 3344 x 1714 pixels.

ajoubertza:
In PR description:
> columns from 2 to 4.

Screen shot shows 5 columns?

bxaia:
> Does it adapt to the number of proxies/components available(dynamic)? or is it statically set?

It is statically set.

bxaia:
> In PR description:
> 
> > columns from 2 to 4.
> 
> Screen shot shows 5 columns?

Sorry it was a typo on the PR description, I will test in depth with different screens and and also test with fewer proxies/components. It is also possible to give both proxies and components 50% height, but this was the initial implementation since there were more proxies than components

ajoubertza:
> It is also possible to give both proxies and components 50% height, but this was the initial implementation since there were more proxies than components

We can use a different ratio - it was just an example.

bngcebetsha:
# Work In Progress
#### On a big secondary screen I see
onBigScreen

#### On the primary screen I see
onMainScreen

Not sure if this can be made to look the same. Will spend some time checking

ajoubertza:
I've updated the description with the new layout from 48fea6e.

app/sensor-list/sensor-list.html
```diff
@@ -1,14 +1,14 @@
...
+        <div layout="column" style="width: 432px">
             <div layout="column" ng-repeat="node in vm.nodes | orderBy:node:true track by $index" class="md-whiteframe-z1"
-                style="margin-bottom: 8px;" ng-class="{'at-least-400-high': node === 'Components' || 'Proxies'}">
+                style="margin-bottom: 8px;" ng-class="{'at-least-200-high': node === 'Components' || 'Proxies'}">
                 <md-toolbar class="md-whiteframe-z1 md-toolbar-tools-medium" layout="row" layout-align="none center">
                     <span>{{node}}</span>
                 </md-toolbar>
                 <div style="overflow: auto" layout-wrap layout="row" layout-align="start">
-                    <div flex="50" ng-class="{'active-sensor-list-resource':vm.resourceSensorsBeingDisplayed === resource.name}"
+                    <div flex="25" ng-class="{'active-sensor-list-resource':vm.resourceSensorsBeingDisplayed === resource.name}"
```
ajoubertza, line 11: For those wondering, the `flex` parameter is a percentage.  With 2 columns, each uses 50%.  With 4 columns, each use 25%.

app/sensor-list/sensor-list.html
```diff
@@ -1,14 +1,14 @@
 
-<div layout="column" flex ng-controller="SensorListCtrl as vm" style="min-height: 825px; min-width: 1000px" ng-keypress="setFilterOnEnter($event, searchText)">
+<div layout="column" flex ng-controller="SensorListCtrl as vm" style="min-height: 600px; min-width: 1000px" ng-keypress="setFilterOnEnter($event, searchText)">
     <div flex layout="row" style="margin: 8px">
-        <div layout="column" style="width: 220px">
+        <div layout="column" style="width: 432px">
```
ajoubertza, line 4: This was updated by inspection.  With two columns, each label had 106 pixels.  `2*106 + 8 = 220`.  With four columns, we use `2*106 + 8 = 432`.

ajoubertza (approved):
LGTM

amakhaba (approved):
LGTM

ajoubertza:
It seems like the problem with the latest dist files (and local via gulp webserver), is related to a change in the [`$mdtoast` API](https://material.angularjs.org/1.2.3/api/service/$mdToast#mdtoast-simple).

Error:
image

Fix:
```diff
diff --git a/app/services/notify-service.js b/app/services/notify-service.js
index 2658071..81f54a8 100644
--- a/app/services/notify-service.js
+++ b/app/services/notify-service.js
@@ -21,7 +21,7 @@
             if (!api.currentToast) {
                 api.currentToast = $mdToast.simple();
                 api.currentToast
-                    .content(message)
+                    .textContent(message)
                     .hideDelay(false) // never hide automatically
                     .position(api.toastPosition);
                 $mdToast.show(api.currentToast);
```

See all the breaking changes for 1.2.0 here:  https://github.com/angular/material/blob/v1.2.2/CHANGELOG.md#breaking-changes.  E.g., we should fix `$mdDialog`, which has a similar change.

ajoubertza (changes requested):
Need to fix the `content` to `textContent` issues

ajoubertza:
It looks like the previous working builds were made with version angular-material 1.1.6.

image

ajoubertza:
> It looks like the previous working builds were made with version angular-material 1.1.6. (This differs from the package.json promise of 1.2.2)

After some unsuccessful testing, with newer packages, we've decided to revert to the older packages (angular 1.6.10, angular-material 1.1.6).

ajoubertza (approved):
LGTM

PR #602, opened 2021-08-31
Set default centre frequency in narrowband mode

bngcebetsha:
We set and display the default narrowband center frequency on L-band. In conjunction with changes in katconfig
#### Before activating a wide-only user product
Screenshot 2021-08-31 at 12 34 56

#### After activating a wide + narrow user product
Screenshot 2021-08-31 at 12 57 06

JIRA: [MT-2566](https://skaafrica.atlassian.net/browse/MT-2566)


app/scheduler/scheduler-home.js
```diff
@@ -354,22 +364,19 @@
 
             vm.subarray.band = band;
             vm.setProduct(product);
-            vm.updateFrequency(label, frequency);
+            vm.updateFrequency(label, frequency, nb_frequency);
```
lanceWilliams, line 367: Is this the only place that the 'updateFrequency' function is used?

ajoubertza (changes requested):
Still needs to be updated to use new key-value definition in katconfig:  `narrowband_defaults`

ajoubertza:
Has this been tested for other bands?  It doesn't look like it would work - we aren't storing the narrowband defaults for each band.

app/scheduler/scheduler-home.js
```diff
@@ -358,14 +370,12 @@
...
-          if (vm.productsWithNarrowBands.includes(vm.subarray.product))
-            defaultNarrowFreq = freq;
-
-          for (var i=0; i<=ObsSchedService.numberOfNarrowBands; i++)
-            vm.setNarrowFrequency(defaultNarrowFreq, i);
+            vm.subarray.requested_rx_centre_frequency = freq;
+            
+            for (var i=0; i<=Object.keys(vm.defaultNarrowBandFreqMap).length; i++){
+                var defaultNarrowFreq = vm.defaultNarrowBandFreqMap[vm.subarray.band]['nb' + i];
+                vm.setNarrowFrequency(defaultNarrowFreq, i);
```
ajoubertza, line 377: Is this loop correct?  The keys for `vm.defaultNarrowBandFreqMap` should be the band letters:  `l`, `s`, `u`, `x`.  That means `i` goes from 0 to 4.  That happens to equal 5 items, which is the number of narrowband instruments we allow.  However the number of bands and number of narrowband instruments are not related.  That is why we used to use the `ObsSchedService.numberOfNarrowBands` value.

Also, the keys in a single band's dict, i.e., in `vm.defaultNarrowBandFreqMap[vm.subarray.band]` would be `nb1`, `nb2`, ..., `nb5` (from `labelledNarrowbandDefaults`).  That means we skip the last one, and try look up a missing item `nb0`.

ajoubertza (changes requested):
The different bands are handled better now, but there is a problem with edge cases.

app/scheduler/scheduler-home.js
amakhaba: This is correct, previously when we were directly obtaining the narroband_freq defaults from the dictionary `defaultNarrowBandFreqMap` that's why we were using `=Object.keys(vm.defaultNarrowBandFreqMap).length`. What you are mentioning is a small detail we missed when handling the new implementation. I will address it.

app/scheduler/scheduler-home.js
```diff
@@ -358,14 +370,12 @@
...
-
-          var defaultNarrowFreq = 0;
-          if (vm.productsWithNarrowBands.includes(vm.subarray.product))
-            defaultNarrowFreq = freq;
-
-          for (var i=0; i<=ObsSchedService.numberOfNarrowBands; i++)
-            vm.setNarrowFrequency(defaultNarrowFreq, i);
+            vm.subarray.requested_rx_centre_frequency = freq;
+            
+            for (var i=0; i<=i<=ObsSchedService.numberOfNarrowBands; i++){
```
ajoubertza, line 375: Two questions:
- Is `i<=i<=ObsSchedService.numberOfNarrowBands` valid JavaScript syntax?
- Why do we start from 0 here, but start from 1 on line 166?

bngcebetsha: big mistake, will fix this now

ajoubertza:
@bngcebetsha Has the latest code, @bngcebetsha, 4e39b6a, been testing for different bands and confirmed to be working as expected?

bngcebetsha:
> @bngcebetsha Has the latest code, [4e39b6a](https://github.com/ska-sa/katgui/commit/4e39b6a30e8bbe33ecca9bd334aa3eb1e53a1f66), been testing for different bands and confirmed to be working as expected?

yes

ajoubertza (approved):

PR #603, opened 2021-09-03
Fix disappearing sky plot targets

ajoubertza:
Previously the targets would only appear for a second after clicking draw.  Now they persist.  The drop-down selectors have also been fixed so that the selected items remain visible.

**Warning:** The positions of the targets are not updated over time.  The azimuth and elevation coordinates are determined when first drawn.  The whole browser page must be reloaded before new coordinates are determined.

Before (_Draw All_ was clicked, but targets disappeared):

image

After (_Draw All_ was clicked):

image

-------
(The JIRA issue below isn't directly related, but the problem was discovered as part of the testing)
JIRA: [P2M-121](https://skaafrica.atlassian.net/browse/P2M-121)


bngcebetsha (approved):
LGTM

PR #608, opened 2021-11-23
Improve time remaining to next global sync

mamkhari:
The GUI indicates the time remaining when the subarray is built, but it isn’t visible on the pages normally visible when running observations.  This means the problem can easily be missed.  This is especially true on RTS that doesn’t have a dedicated operator.

This PR does the following:

- Shows the NGS time on all 3 subarray pages: `set-up subarray`, `manage schedule blocks`, and `observations`
- Ensures that NGS time goes red when 0 hours left

These improvements are necessary so that the user can have a view of how long they have before they have to run
global sync. Saves telescope time as they will know this before running observations and not have them archive on SDP

The following screenshots show the NGS time on all 3 subarray pages, and red font when NGS time is 0 hours.

Screenshot from 2021-11-23 20-15-59
Screenshot from 2021-11-23 20-17-01
Screenshot from 2021-11-23 20-17-20

JIRA: [MT-1877](https://skaafrica.atlassian.net/browse/MT-1877)


app/scheduler/observations/observations-detail.html
```diff
@@ -24,14 +24,21 @@
...
+                        </span>
+                    </md-button>
                 </md-menu>
 
                 <span flex></span>
 
                 <md-menu style="margin: 0 8px; margin-bottom: 0; padding: 0; position: relative">
                     <span md-ink-ripple style="margin: 0; padding: 0px" ng-click="$mdMenu.open($event)">
                         <span class="fa fa-chevron-down subarray-select-chevron" style="top: 7px; opacity: 1.0"></span>
-                        <span style="margin-left: 16px">MODE: <b>{{vm.subarray.mode}}</b></span> </span>
```
bngcebetsha, line 34: good, thanks for picking up redundant tag

amakhaba: Are we sure its redundant ?

bngcebetsha: I am sure, I stand to be corrected with an argument

tockards:
> I am still wondering if we can optimize the changes by only calculating `vm.minGlobalSyncTime` in one module(say `scheduler-home.js`) and import it in the places it must be used. Instead of re-calculating it on all pages.

The one issue with this is that how do we guarantee that we when we call upon this value it is the most recent value of the global sync time?  I am under the impression that it is a dynamic value and changes over a period of time hence `minGlobalSyncTime` part in the name.

bngcebetsha (approved):

tockards (approved):
LGTM and it worked in testing. 
You might just need @ofaletse or @bxaia to advise on the correct pattern for the Javascript as we have it in multiple places.

mmphego:
@mamkhari you forgot to create the dists for katgui. 
Please create another PR to address that.

mamkhari:
> @mamkhari you forgot to create the dists for katgui. Please create another PR to address that.

Thanks, @mmphego .

PR #609, opened 2021-11-24
Hotfix V30: Improve UX for GUI remaining time to next global sync

mamkhari:
The scheduler page on katGUI indicates the time (in hours) remaining before the next global sync (via the `NGS --- Hrs` display). This display is not visible on the `Manage Schedule Blocks` and `Observations` pages.  This means the deadline for global sync can easily be missed while switching from page to page.  This is especially true on RTS system that doesn’t have a dedicated Operator.

This PR does the following:

- Shows the `NGS` display on all 3 scheduler pages: `set-up subarray`, `manage schedule blocks`, and `observations`
- Ensures that `NGS` display goes red when `0` hours left

These improvements are necessary so that the user can have a view of how long they have before they have to run
global sync. Saves telescope time as they will know this before running observations and not have them archive on SDP

The following screenshots show the NGS time on all 3 subarray pages, and red font when NGS time is 0 hours.

JIRA: [MT-1877](https://skaafrica.atlassian.net/browse/MT-1877)


mmphego (approved):
Looks good, please don't forget the dist files

mmphego (approved):
Looks good

bngcebetsha (approved):

PR #611, opened 2021-11-25
add step to check if dist file was updated

tockards:
When js, css and html files are updated and a new dist is not built we need to notify the user somehow that the PR is incomplete.
 
This PR adds a step to Jenkins Job that checks whether the dist files were updated by
 - comparing the current changes to the master branch.
 - checks if the `dist` directory was updated using `grep`.
 - exits if it was not updated. 

Please note that is a simple and naive way of checking if the dist files were updated.


tockards:
closing this to test it in jenkins and possibly move it to our `lint_diff.sh`

Jenkinsfile
```diff
@@ -17,6 +17,12 @@ pipeline {
                 ])
             }
         }
+        
+        stage ('Check if dist file were built') {
```
mmphego, line 21: Nice @tockards, this would be very useful. Perhaps fail early instead of echoing. We should mob it!

PR #615, opened 2021-12-15
Improve visibility of additional tags on the User Logs page

mamkhari:
Some of the tag names in the display that appear in the `Additional Tags` drop-down menu, do not expand to indicate what those tags are. So, it is difficult for the operators to be able to select the correct tag.

This PR does the following:
- Increase the width of the drop-down menu to make the contents more readable

The following screenshots show how the drop-down menu was before and after this change.

Screenshot from 2021-12-15 06-29-50
Screenshot from 2021-12-15 06-34-20

JIRA: [MT-2710](https://skaafrica.atlassian.net/browse/MT-2710)


mamkhari:
Attached is the screenshot showing how the drop-down  box is aligned to the width of the input box (the red box) 

Screenshot from 2021-12-15 14-01-21

shoosen (commented):
Visually, this looks great, but I feel we will still need a tooltip for those compound tags that may be longer than this drop-down menu.

bngcebetsha:
Looks really nice @mamkhari, thank you @tockards and @shoosen . I was wondering if it will be too much trouble to get the search bar(highlighted orange) to be  the same size as the drop-down that comes up when one starts typing.

bngcebetsha (approved):
@mamkhari  please rebuild the dist files for good measure before you merge.

tockards (approved):

PR #618, opened 2022-01-24
Update the rsc-aggregate-color for the receiver ready icon on katgui

bngcebetsha:
In this PR, we update the `isReceiverReady()`  icon from `WARN`(yellow) to `ERROR`(red) when the receiver is NOT ready (aggregate has a value of false). The only difference between this and previous functionality is the color of the icon and what status it reflects.

#### Before
Screenshot 2022-01-19 at 15 29 47

#### After
Screenshot 2022-01-24 at 14 56 24

Screenshot 2022-01-24 at 14 55 49

JIRA: [MT-2843](https://skaafrica.atlassian.net/browse/MT-2843)


amakhaba:
Please attach screenshots @bngcebetsha when toggling the error status

PR #627, opened 2022-05-20
Add CI stage for katgui to auto build dist files after a PR has been approved and ready for merge

amakhaba:
The katgui Jenkins job already builds the dist files as a check.  However, developers must still manually build, commit and push the dist files from their local development environment to create a new release.  This is tedious and error-prone.  

This PR adds a stage in the Jenkins file to automatically build, commit and push dist files when a Pull request has been approved and ready for merge.

distfilescommit

JIRA: [1843](https://skaafrica.atlassian.net/browse/1843)

PR #628, opened 2022-05-25
Run jenkins stage when branch is not master

amakhaba:
In an effort to avoid building and committing dist files when the branch is in master, this PR adds a condition that specifies that the step should only run when the branch is not master. This work is linked to [PR-627](https://github.com/ska-sa/katgui/pull/627)

JIRA: [1843](https://skaafrica.atlassian.net/browse/1843)


Jenkinsfile
```diff
@@ -54,6 +54,10 @@ pipeline {
 
         stage("Commit & Push dist files"){
             when {
```
mmphego, line 56: Only one `when` clause is allowed per `stage` as per [doc](https://www.jenkins.io/doc/book/pipeline/syntax/#when). You need two `stages`s excluding each other by `when` clauses to achieve the if-else behaviour you want. Alternatively, you can use `anyOf` which will skip the remaining step as soon as the first "true" condition is found.

```groovy
when {
    anyOf {
        not {
            branch 'master'
        }

        expression {
            env.PR_APPROVED == "update"
        }
    }
}
```

mmphego (changes requested):

Jenkinsfile
amakhaba: @mmphego using the above-mentioned document as a reference, by checking the examples page you see that `when` block can contain more than one condition, and in that instance "If the when directive contains more than one condition, all the child conditions must return true for the stage to execute" In our particular case we do want the stage to execute when we not in master and also when the PR is approved

mmphego: Agreed!

mmphego (approved):

PR #630, opened 2022-05-29
Plot two y-axis on sensor graph

bngcebetsha:
As part of the much larger feature described in https://skaafrica.atlassian.net/browse/MKT-148, where we want to eventually plot both numeric and discrete sensors on the same set of axes.

In this PR we place two y-axis on the left and on the right. See screenshot for a better view.

#### Numeric
Screenshot 2022-05-29 at 13 23 46

#### Discrete
Screenshot 2022-05-29 at 13 23 15

JIRA: [MT-2990](https://skaafrica.atlassian.net/browse/MT-2990)


bngcebetsha:
We still need to toggle the y-axis on the right when there is only one sensor being plotted. Still searching the codebase for a variable that indicates this. For now, consider this a draft PR

PR #635, opened 2022-07-20
Using `gulp-uglify-es` module instead of `gulp-uglify`

mmphego:
This PR does the following:

- Replaces `gulp-uglify` module with a pinned version of `gulp-uglify-es`, 
UglifyJS [does not currently support](https://github.com/mishoo/UglifyJS2/issues/448) EcmaScript 6 structures [like classes](https://github.com/mishoo/UglifyJS2/issues/659) thus resulting in an error during build. Note `gulp-uglify-es` is no longer being maintained, consider `gulp-terser`.

## Error during build
image

## Results
image


bngcebetsha (approved):
LGTM

PR #636, opened 2022-08-03
Finalise Plotting two y-axis on the same time series

tockards:
In this PR we finalise the feature as on the PR title. We had to rework the codebase in order to be able to export the URL to an already plotted set of data. This helps the user save time and for the system to make fewer request to the data store, as would be the case if they replotted when they want to share their plots.

JIRA: [MT-3056](https://skaafrica.atlassian.net/browse/MT-3057)


bngcebetsha (commented):
when searching for sensors, both numeric and discrete appear on the list. Plotting them is not working for some reason.

Screenshot 2022-08-03 at 10 28 40
Screenshot 2022-08-03 at 10 28 51

app/d3/multi-line-chart.js
```diff
@@ -560,9 +692,47 @@ angular.module('katGui.d3')
...
+                                        return d.min_value;
+                                    }
+                                    var parsedValue = parseFloat(d.value || (d.avg_value || 0));
+                                    if (isNaN(parsedValue)) {
+                                        return d.value || (d.avg_value || 0);
+                                    }
+                                    return parsedValue;
+                                });
+                            }),
+                            d3.max(scope.nestedData.filter(d => d.type === "type"), function(sensors) {
```
tockards, line 719: use `discrete` instead of `type`

app/d3/multi-line-chart.js
```diff
@@ -560,9 +692,47 @@ angular.module('katGui.d3')
...
+                }
+                if (scope.options.useFixedYAxis && !scope.options.yAxisDiscreteValues) {
+                    if (scope.options.discreteSensors) {
+                        yDiscrete.domain([scope.yMinDiscrete, scope.yMaxDiscrete]);
+                    }
+                } else if (!scope.options.yAxisDiscreteValues) {
+                    var yExtent = [0, 0];
+                    if (scope.nestedData.length > 0) {
+                        yExtent = [
+                            d3.min(scope.nestedData.filter(d => d.type === "type"), function(sensors) {
```
bngcebetsha, line 707: "type" should be "discrete"

bngcebetsha (commented):
Perhaps as a last thing to loo into:
- when changing from numeric to discrete, upon plotting - the list refreshes briefly, so we want to prevent that.
- Not sure if we can make both numeric and discrete sensors appear on the zoom-in window at the bottom.

app/sensor-graph/sensor-graph.js
```diff
@@ -401,9 +426,20 @@
...
+                hideContextZoom: !vm.showContextZoom,
+                useFixedYAxis: vm.useFixedYAxis,
+                useFixedXAxis: vm.useFixedXAxis,
+                xAxisValues: [vm.sensorStartDatetime, vm.sensorEndDatetime],
+                discreteSensors: vm.searchDiscrete
+            });
+        }
 
         vm.sensorTypeChanged = function () {
-            vm.clearData();
```
bngcebetsha, line 406: this part essentially makes `vm.sensorTypeChanged` to be exactly the same as `vm.sensorTypeChangedUpdate`

lanceWilliams (approved):
We should monitor closely after deployment but it is looking really good for what I have tested at least. File download seems fully functional as well. Let's try it.

bngcebetsha (approved):

